PR #2, opened 2020-01-30
finished part 1

camillalof:
Im very confused this week, but I think I somehow managed to learn some basics with the help of Jennie. So I am proud that I have some type of project to deploy. 

I think I could of made this better by adding more path parameters. I didnt really know how to write more of theese and was worried to brake the code I succeded in if I would try. But if there is a way to do that I would of love to get feedback on that.

Thanks alot!

### DIFF
--- a/server.js
+++ b/server.js
@@ -1,15 +1,21 @@
 import express from 'express'
 import bodyParser from 'body-parser'
 import cors from 'cors'
+import booksData from './data/books.json'
 
-// If you're using one of our datasets, uncomment the appropriate import below
-// to get started!
-// 
-// import goldenGlobesData from './data/golden-globes.json'
-// import avocadoSalesData from './data/avocado-sales.json'
-// import booksData from './data/books.json'
-// import netflixData from './data/netflix-titles.json'
-// import topMusicData from './data/top-music.json'
+/* Sample from API
+
+  "bookID": 1,
+  "title": "Harry Potter and the Half-Blood Prince (Harry Potter  #6)",
+  "authors": "J.K. Rowling-Mary GrandPré",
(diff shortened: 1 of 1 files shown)


JennieDalgren:
Can you add a link to your deployed project?

camillalof:
https://express-deployment-camilla.herokuapp.com/

server.js
```diff
@@ -27,7 +33,42 @@ app.get('/', (req, res) => {
   res.send('Hello world')
 })
 
+// Path variable/path parameter
+app.get('/books/:id', (req, res) => {
+  const bookId = req.params.id;
+  const book = booksData.filter(item => item.bookID === +bookId);
```
JennieDalgren: since you are looking for only one book here, there will only be one book with the specific id, you should use the find() method instead of filter(). `find()` returns a single object whereas `filter()` returns an array.

server.js
```diff
@@ -27,7 +33,42 @@ app.get('/', (req, res) => {
...
+  console.log('hej', searchString)  
+    filteredBooks = filteredBooks.filter(item => {
+      const itemTitle = item.title.toString().toLowerCase();
+      const itemAuthors = item.authors.toString().toLowerCase();
+      //the text in the searchfield will always be a string (det som står i sökfältet kommer alltid vara en string, när Damien gjorde ett + i texten så var det för att göra sökfältet till en int.)
+      console.log(itemTitle)
+      return itemTitle.includes(searchString.toLowerCase()) ||
+        itemAuthors.includes(searchString.toLowerCase());
+    });
+  }
```
JennieDalgren: nice job! Also, good explanation of the code ⭐️

PR #4, opened 2020-02-02
Project express api – Emmie

eilsson:
**Link to project on Heroku**
https://express-entertainment.herokuapp.com/

When I send a message when an item is not found, should I include the status code as an entry as well as the message?

When I filter countries I divided up my code a bit, so instead of
```  
if (countrySearchString) {
  filteredShows = shows.filter((show) => {
    return show.country.toLowerCase().includes(countrySearchString.toLowerCase())
  })
}
```

I wrote it like this
```
if (countrySearchString) {
  filteredShows = shows.filter((show) => {
    const showCountries = show.country.toLowerCase()
    const searchString = countrySearchString.toLowerCase()

    return showCountries.includes(searchString)
  })
}
```
I think the latter is easier to read, but maybe that's just me?


server.js
```diff
@@ -1,33 +1,98 @@
...
+  }
+
+  res.json(filteredShows)
+})
+
+app.get('/shows/:show', (req, res) => {
+  const showId = req.params.show
+
+  const shows = netflixData.filter((item) => item.type === "TV Show")
+  const show = shows.find((item) => item.show_id === +showId)
```
JennieDalgren: great that you're using find here, which is only returning one item

PR #5, opened 2020-02-02
Project-express-api - Susanna Sjöberg

SanSjo:
Hi ,
This week has been very interesting and fun to learn backend. It's been hard to grasp but I got more clear about it after every codealong. 

I tried to add frontend to it, but all my code crashed when I tried to add the react-starter to the file, I got it back though and tried again but it didn't work. I didn't have time to solve it, so I decided to go back to my backend code for this project. But will try to add the frontend another time. 

Here is my deployed site: http://project-express-api.herokuapp.com/api

### DIFF
--- a/server.js
+++ b/server.js
@@ -1,33 +1,114 @@
-import express from 'express'
-import bodyParser from 'body-parser'
-import cors from 'cors'
+import express from 'express';
+import bodyParser from 'body-parser';
+import cors from 'cors';
 
 // If you're using one of our datasets, uncomment the appropriate import below
 // to get started!
-// 
-// import goldenGlobesData from './data/golden-globes.json'
+//
+// import goldenGlobesData from './data/golden-globes.json';
 // import avocadoSalesData from './data/avocado-sales.json'
-// import booksData from './data/books.json'
-// import netflixData from './data/netflix-titles.json'
+import booksData from './data/books.json';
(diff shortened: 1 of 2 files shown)


server.js
```diff
@@ -1,33 +1,71 @@
...
 // Add middlewares to enable cors and json body parsing
-app.use(cors())
-app.use(bodyParser.json())
+app.use(cors());
+app.use(bodyParser.json());
+
+// Two routes here
+
+//First route filter once on multiple fields based on title OR country OR desription OR genre
+app.get('/api', (req, res) => {
```
JennieDalgren: this endpoint should be more descriptive than `/api` so if the api is about books, call it /books for example

server.js
```diff
@@ -1,33 +1,71 @@
...
+      const itemTitle = item.title.toString();
+      const itemCountry = item.country.toString();
+      const itemDescription = item.description.toString();
+      const itemGenre = item.listed_in.toString();
+      return (
+        itemTitle.includes(searchString.toLowerCase()) ||
+        itemCountry.includes(searchString) ||
+        itemDescription.includes(searchString) ||
+        itemGenre.includes(searchString)
+      );
```
JennieDalgren: nice to be able to search freely regardless if its a country or title and so on...

PR #6, opened 2020-02-02
Project 17 Build an API using Express

Angelicabrodin:
I felt that it was really hard to move to back end, again every week goes so fast and you really do not have time to let everything sink in unfortunately. 

I had so much problems with my server, and it was giving a lot of errors, and then I also had problems with heruko, I do not know if this is because I tried changing the port nr so many times to try and see if that was what made my server not working. And in the end, I still have a error when running npm run dev but heruko finally works so that is good. 

I think it is cool when you think about how much you can create, and I wish I had so much more time to experiment, I will hopefully go back to this in the future and try it out a bit more!

### DIFF
--- a/server.js
+++ b/server.js
@@ -1,10 +1,12 @@
 import express from 'express'
 import bodyParser from 'body-parser'
 import cors from 'cors'
+import goldenGlobesData from './data/golden-globes.json'
+
+console.log(goldenGlobesData.length)
 
 // If you're using one of our datasets, uncomment the appropriate import below
 // to get started!
-// 
 // import goldenGlobesData from './data/golden-globes.json'
 // import avocadoSalesData from './data/avocado-sales.json'
 // import booksData from './data/books.json'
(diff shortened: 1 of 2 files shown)


JennieDalgren:
please add link to deployed project

Angelicabrodin:
> please add link to deployed project

@JennieDalgren I thought it was enough with heruko? do I need to deploy it on netlify as well?

server.js
```diff
@@ -2,19 +2,23 @@ import express from 'express'
 import bodyParser from 'body-parser'
 import cors from 'cors'
 
+import goldenGlobesData from './data/golden-globes.json'
+import netflixData from './data/netflix-titles.json'
+
+console.log(goldenGlobesData.length)
+console.log(netflixData.length)
```
JennieDalgren: remove this

server.js
```diff
@@ -27,6 +31,63 @@ app.get('/', (req, res) => {
...
+//   "type": "TV Show"
+// },
+
+// Netflix - Titles
+
+app.get('/movies', (req, res) => {
+  res.json(netflixData)
+})
+
+app.get('/releaseyear/:year', (req, res) => {
```
JennieDalgren: again here, call this movies/:year or make it an optional query in the movies endpoint

PR #7, opened 2020-02-02
PR - Tiago Ferreira

tavferreira:
Link to my deployed project:
Backend: https://tavferreira-bookstore-api.herokuapp.com/books?page=0
Frontend: https://dazzling-kepler-13501e.netlify.com/

Initially the project seemed easy, but after adding frontend to it I've realized that the API was not working as expected in many aspects. Also, by adding the frontend to it, I've realized that that took a lot more time to implement, even and ugly UI as this one.

Things I'd like to have clarified or explained in more detail: It was hard to deploy the frontend to netlify without the netlify.toml and structure. It took me quite a while to make it work.


server.js
```diff
@@ -27,7 +19,54 @@ app.get('/', (req, res) => {
...
+const PER_PAGE = 10
+
+app.get('/books', (req, res) => {
+  const { page } = req.query
+  const startIndex = PER_PAGE * +page
+  let booksData = books
+
+  const { lang } = req.query
+  const { title } = req.query
+  const { rating } = req.query
```
idlefingers: Great with destructuring! These should all be on one line:

```js
const { lang, title, rating } = req.query
```

server.js
```diff
@@ -27,7 +19,54 @@ app.get('/', (req, res) => {
...
+
+
+  if (rating) {
+    if (rating === 'asc') {
+      booksData.sort(function (a, b) { return a.average_rating - b.average_rating })
```
idlefingers: Nice! Don't forget to stick with es6, though:

```js
booksData.sort((a, b) => a. average_rating - b. average_rating)
```

server.js
```diff
@@ -27,7 +19,54 @@ app.get('/', (req, res) => {
...
+  }
+
+  if (lang) {
+    booksData = booksData.filter(book => book.language_code === lang)
+  }
+
+  if (title) {
+    booksData = booksData.filter(book => book.title.toString().toLowerCase().includes(title.toLowerCase()))
+  }
+  const length = booksData.length
```
idlefingers: There's no real issue with this, but I don't think this variable makes the code any more readable - just makes the line count longer. I'd just use `booksData.length` in the totalPages line further down.

server.js
```diff
@@ -27,7 +19,54 @@ app.get('/', (req, res) => {
...
+    totalPages: Math.floor(length / PER_PAGE),
+    currentPage: +page,
+    booksData
+  })
+})
+
+app.get('/books/:id', (req, res) => {
+  const id = req.params.id
+
+  const book = books.filter((book) => book.bookID === +id)
```

server.js
```diff
@@ -27,7 +19,54 @@ app.get('/', (req, res) => {
...
+
+  res.send(book)
```
idlefingers: I'm not actually sure in this instance if there's any difference since express is smart enough to see you're sending a JS object here and so it makes it a json response, but you should aim to be explicit and use `res.json(...)`

server.js
tavferreira: I have this since it is possible to query by language or title and the size of the array would change after that.
If I use booksData.length after the slice the length would be incorrect and if I use books.length the number of pages could be wrong if there was any query filtering the data.
Any suggestions?

PR #8, opened 2020-02-02
Project 17 - Fredrik Tibbling

codebyfredrik:
URL to custom Netflix API documentation: https://express-netflix-api.herokuapp.com/

This has been a super interesting project to work on. From a learning experience perspective I think this has been one of my best weeks so far. I've learned so much. I completed the basic requirements plus the stretch goals early on in the week so I came up with a couple of my own stretch goals that I researched and implemented in my project. I'm really happy with the end result. 

Apart from the GET request method I also decided to implement POST, PUT and DELETE request methods in my project. I have also put in time and effort to handle relevant status codes in a graceful way.

With everything I had going on this week I didn't have time to hookup a frontend to the backend. That's on my to-do list as soon as I have some free time.

**Things I'd like to have clarified or explained in more detail:** -


routes.js
```diff
@@ -0,0 +1,327 @@
...
+import path from 'path';
+import Joi from 'joi';
+import { requireJsonContent } from './middleware/index';
+import netflixData from './data/netflix-titles.json';
+
+const router = express.Router();
+const ITEMS_PER_PAGE = 20;
+let remainingPages = null;
+let totalPages = null;
+let movies = netflixData;
```
idlefingers: If you're going to use `let` variables like this, they should be defined in the route - defining them in the global scope like this will lead to strange bugs or confusion if you're not careful

routes.js
```diff
@@ -0,0 +1,327 @@
...
+});
+
+router.get('/movies', (req, res) => {
+  const queryPage = +req.query.page;
+  const queryYear = +req.query.year;
+  const queryDuration = +req.query.duration;
+  const queryActor = req.query.actor;
+
+  // Local copy of movies
+  let filteredMovies = movies;
```
idlefingers: I think it'd make more sense to just use `netflixData` here. All the intermediary variables make it more confusing

routes.js
```diff
@@ -0,0 +1,327 @@
...
+          return;
+        }
+      });
+
+      // Return current movie if actor is found
+      if (result.length > 0) {
+        return movie;
+      }
+    });
+  }
```
idlefingers: Try to avoid having too much logic going on in one place. This filter is pretty difficult to follow, with a filter in a filter, and returning the item rather than a boolean - it makes it all a bit difficult to mentally parse. When the logic starts getting thick like this, try to push yourself to break things up a little more and make use of functions to encapsulate and self-document what's going on. 👍

routes.js
```diff
@@ -0,0 +1,327 @@
...
+
+  // Query - Page
+  if (queryPage) {
+    const startIndex = queryPage * ITEMS_PER_PAGE - ITEMS_PER_PAGE; // Page 1 starts at index 0
+    const endIndex = startIndex + ITEMS_PER_PAGE;
+    totalPages = Math.ceil(filteredMovies.length / ITEMS_PER_PAGE);
+    remainingPages = totalPages - queryPage;
+    filteredMovies = filteredMovies.slice(startIndex, endIndex);
+  }
```
idlefingers: I think this should be the default behavior - not returning all items. If this were a real API using a database, it would be very heavy to return all items all the time, so it's good to limit by default.

PR #9, opened 2020-02-02
Project 17 - Sofie Nyblad

nyblad:
**Live link:**
Backend: https://nyblad-express-api.herokuapp.com
Frontend: https://snack.expo.io/@nyblad/nyblad-express-api (probably not best practise on how I filter the data)

Up and down, to set up the basics felt easy but to get the pagination to work with different endpoints and queries was harder. Left some outcommented stuff since I still have an issue (explained below) and not sure what way to go yet.

**Questions:**
-  I still can't get the filteredPages correct when searching for a title, when I apply two quearies like https://nyblad-express-api.herokuapp.com/shows?title=king&&page=0 I want to get the total pages of the search result, not all the pages.

### DIFF
--- a/server.js
+++ b/server.js
@@ -16,7 +16,7 @@ app.use(bodyParser.json())
 
 // ROOT ENDPOINT
 app.get('/', (req, res) => {
-  res.send('API for Netflix shows, available endpoints: /shows, /:id, /types/:type, ?page=X for pagination')
+  res.send('API for Netflix shows, available endpoints: /shows -> /:id, /movies, /tv-shows, ?title=X for search and ?page=X for pagination')
 })
 
 // GET ALL DATA WITH PAGINATION OR FILTERED WITH TITLE IN QUERY
--- a/server.js
+++ b/server.js
@@ -22,14 +22,23 @@ app.get('/', (req, res) => {
 
 // GET ALL DATA WITH PAGINATION OR FILTERED WITH TITLE IN QUERY
 app.get('/shows', (req, res) => {
-  // Variable for qeury parameter
+  // For qeury page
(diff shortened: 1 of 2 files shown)


server.js
```diff
@@ -1,33 +1,106 @@
...
+
+  // if (typeSearchString) {
+  //   filteredShows = filteredShows.filter((item) => item.type.toLowerCase().replace(' ', '-') === typeSearchString.toLowerCase())
+  // }
+
+  if (page) {
+    const startIndex = PER_PAGE * +page
+    // StartIndex and the upcoming shows updates when increasing page number in endpoint
+    filteredShows = filteredShows.slice(startIndex, startIndex + PER_PAGE)
+  }
```
idlefingers: I think it'd be nice to have this pagination as a default - returning so many responses from the API as a default is pretty heavy, and no frontend is going to be ready for that many results all at once - better to break them up!

server.js
```diff
@@ -1,33 +1,106 @@
...
+
+  // Return data and total pages of data
+  res.json({
+    totalPages: Math.floor(netflixData.length / PER_PAGE),
+    filteredPages: Math.floor(netflixData.length / PER_PAGE),
```
idlefingers: What's the difference between total and filtered here?

nyblad: Right now, no difference. I want filteredPages to show the total pages of the filtered shows (not the total total). But it went nuts if i put it like filteredShows.length, then it says there just 2 filteredPages for https://nyblad-express-api.herokuapp.com/shows?page=0 ?

nyblad: Ah, finally got it to work in the backend.

PR #10, opened 2020-02-02
project-express-api by Agnes

Agnes94:
Link to the deployed project: https://express-api-project.herokuapp.com/

### DIFF
--- a/server.js
+++ b/server.js
@@ -1,13 +1,13 @@
 import express from 'express'
 import bodyParser from 'body-parser'
 import cors from 'cors'
+import books from './data/books.json'
 
 // If you're using one of our datasets, uncomment the appropriate import below
 // to get started!
 // 
 // import goldenGlobesData from './data/golden-globes.json'
 // import avocadoSalesData from './data/avocado-sales.json'
-// import booksData from './data/books.json'
 // import netflixData from './data/netflix-titles.json'
 // import topMusicData from './data/top-music.json'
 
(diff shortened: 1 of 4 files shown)


server.js
```diff
@@ -27,6 +27,34 @@ app.get('/', (req, res) => {
   res.send('Hello world')
 })
 
+// Returning all the books (objects) from the array 
+app.get('/books', (req, res) => {
+  const { page } = req.query
+  const startIndex = 20 * +page
```
idlefingers: It would be nice with a default here. If I call the API without a `page` query param, I don't get any results, but if I pass `page=0` then I get the first page. You could use `||` (or) to set a default if `+page` is `NaN`:

```
const startIndex = 20 * +page || 0
```

server.js
```diff
@@ -27,6 +27,34 @@ app.get('/', (req, res) => {
...
+  res.json(books.slice(startIndex, startIndex + 20))
+})
+
+//request.params is an object containing properties to the named route
+// Return an individual book, using :id as a placeholder for the bookID number
+app.get('/books/:id', (req, res) => {
+  const id = req.params.id
+  const bookId = books.filter((item) => item.bookID === +id) // + means turning string into a number
```

server.js
```diff
@@ -27,6 +27,34 @@ app.get('/', (req, res) => {
...
+  res.json(bookId)
```
idlefingers: I think this line can be removed since the response is sent in the condition below

server.js
```diff
@@ -27,6 +27,34 @@ app.get('/', (req, res) => {
...
+  if (bookId) {
+    res.json(bookId)
+  } else {
+    res.status(404).json({ message: 'No books found with this id.' })
+  }
+})
+
+// Returning the books with a specific language code
+app.get('/books/language/:language', (req, res) => {
```
idlefingers: Nice! 😎

PR #11, opened 2020-02-02
API end points completed

Lilivet:
This is my Heroku site:

https://building-api-using-express.herokuapp.com/

These are my end-points:

https://building-api-using-express.herokuapp.com/catalogue
https://building-api-using-express.herokuapp.com/type/movie
https://building-api-using-express.herokuapp.com/type/Movie
https://building-api-using-express.herokuapp.com/type/tvshow
https://building-api-using-express.herokuapp.com/type/tv show
https://building-api-using-express.herokuapp.com/type/TV show
https://building-api-using-express.herokuapp.com/type/show
https://building-api-using-express.herokuapp.com/type/TVshow
https://building-api-using-express.herokuapp.com/year/2019?type=show

### DIFF
--- a/server.js
+++ b/server.js
@@ -1,6 +1,7 @@
 import express from 'express'
 import bodyParser from 'body-parser'
 import cors from 'cors'
+import netflixData from './data/netflix-titles.json'
 
 // If you're using one of our datasets, uncomment the appropriate import below
 // to get started!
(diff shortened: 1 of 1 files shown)


server.js
```diff
@@ -22,7 +23,38 @@ const app = express()
...
 
-// Start defining your routes here
+// Start defining your RESTful end points/routes here
+app.get('/catalogue', (req, res) => {
+  res.json(netflixData)
+})
+
+app.get('/type/:type', (req, res) => {
+  let type = req.params.type.toLowerCase().replace(/[-_ ]/, "")
+  if (type === "tvshow" || type === "show" || type === "tv") {
```
idlefingers: Nice! Just a little tip, you can make use of `Array.includes` to do a lot of checks like this, if you end up with a big list:

```
if (["tvshow", "show", "tv"].includes(type)) {
  ...
}
```

server.js
```diff
@@ -22,7 +23,38 @@ const app = express()
...
+  let catalogueYear = netflixData.filter((item) => item.release_year === +year)
+
+  if (type) {
+    type = type.toLowerCase().replace(/[-_]/, "")
+    if (type === "tvshow" || type === "show" || type === "tv") {
+      type = "TV Show"
+    } else if (type === "movie") {
+      type = "Movie"
+    }
+    catalogueYear = catalogueYear.filter((item) => item.type === type)
```
idlefingers: Just an idea, since this logic is the same as the one in the other endpoint, but using different data, you could abstract the logic a bit to make it reusable. For example, if you made a function like this:

```
const dataByType = (unformattedType, data) => {
  type = unformattedType.toLowerCase().replace(/[-_]/, "")
  if (type === "tvshow" || type === "show" || type === "tv") {
    type = "TV Show"
  } else if (type === "movie") {
    type = "Movie"
  }

  return data.filter((item) => item.type === type)
}
```

Then you could use it in this endpoint:

```
if (type) {
   res.json(dataByType(type, catalogueYear))
} else {
  res.json(catalogueYear)
}
```

And the same approach in the `/type/:type` endpoint above:

```
res.json(dataByType(type, netflixData))
```

server.js
```diff
@@ -22,7 +23,38 @@ const app = express()
 app.use(cors())
 app.use(bodyParser.json())
 
-// Start defining your routes here
+// Start defining your RESTful end points/routes here
+app.get('/catalogue', (req, res) => {
+  res.json(netflixData)
+})
+
+app.get('/type/:type', (req, res) => {
```
idlefingers: Don't forget to use RESTful routes. The route should be the pluralised version of the data you're returning. So if this is returning a 'Show', then this endpoint should be `/shows` and then using a query parameter for the type. So ending up with an endpoint like this: `/shows` - returns everything, `/shows?type=movie` - returns only movies

Lilivet:
Thanks Demien!

I will follow your advice to improve my code!

PR #12, opened 2020-02-02
Project 17 - Nasim Mahzoun

Nasimmhn:
Link to my deployed project:
https://api-express-by-nasim.herokuapp.com 

In the beginning of the project, it felt very confusing. I had problem with "should it be params or query"!!! But after the code alongs and googling, it became more clear and felt quite comfortable.
I chose to do a bit of frontend, BUT since it didnt work with react, I went for <% EJS %> Embedded JavaScript templating. It was totally new for me, but thanks to dear Google, it turned out well. I hope that is ok with you too.

Coming...


server.js
```diff
@@ -1,33 +1,107 @@
...
+
+  res.render('pages/show', {
+    show: show,
+  });
+});
+
+
+
+// http://localhost:8080/api/netflix?country=nigeria&year=2018
+app.get('/api/netflix', (req, res) => {
```
idlefingers: Don't forget to use RESTful routes. The route should be the pluralised version of the data you're returning. So if this is returning a 'Show', then this endpoint should be `/shows` and then using a query parameter for the type. So ending up with an endpoint like this: `/shows` - returns everything, `/shows?year=2019` - returns only shows from 2019

PR #13, opened 2020-02-03
w17

tehvicke:
https://express-deployment-vd.herokuapp.com/

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,6 +15,7 @@
     "body-parser": "^1.19.0",
     "cors": "^2.8.5",
     "express": "^4.17.1",
+    "express-list-endpoints": "^4.0.1",
     "nodemon": "^2.0.2"
   },
   "devDependencies": {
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -15,6 +15,7 @@
     "body-parser": "^1.19.0",
     "cors": "^2.8.5",
     "express": "^4.17.1",
+    "express-list-endpoints": "^4.0.1",
```
idlefingers: 😎

server.js
```diff
@@ -22,11 +27,67 @@ const app = express()
...
+  
+  res.send(listEndpoints(app))
+})
+
+app.get('/netflix-titles/', (req, res) => {
+  let results = []
+
+   /** Fetch query params */
+  const actor = req.query.actor
+  const page = req.query.page || 1
```
idlefingers: You could use destructuring here, and get the default set at the same time:

```
const { actor, page = 1 } = req.query
```

PR #14, opened 2020-02-03
Week 17

elol42:
https://express-api-netflix.herokuapp.com/

It was pretty easy to get a grip around, but i think i need to go a bit deeper to understand how to really create a good "ground"...

### DIFF
--- a/server.js
+++ b/server.js
@@ -1,15 +1,7 @@
 import express from 'express'
 import bodyParser from 'body-parser'
 import cors from 'cors'
-
-// If you're using one of our datasets, uncomment the appropriate import below
-// to get started!
-// 
-// import goldenGlobesData from './data/golden-globes.json'
-// import avocadoSalesData from './data/avocado-sales.json'
-// import booksData from './data/books.json'
-// import netflixData from './data/netflix-titles.json'
-// import topMusicData from './data/top-music.json'
+import netflixData from './data/netflix-titles.json'
 
 // Defines the port the app will run on. Defaults to 8080, but can be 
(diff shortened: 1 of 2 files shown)


server.js
```diff
@@ -24,7 +16,41 @@ app.use(bodyParser.json())
...
+  res.send('Find Netflix data')
+})
+
+// Search the whole data base
+app.get('/movies', (req, res) => {
+  res.json(netflixData)
+})
+
+// Search on country and type Movie or TV Show
+app.get('/country/:country', (req, res) => {
```
idlefingers: I think for this to follow the rules of REST, since the data it returns is an array of movies, this logic should be in the `/movies` endpoint. Then you could use a query param to filter, like `/movies?country=sweden`.

server.js
```diff
@@ -24,7 +16,41 @@ app.use(bodyParser.json())
...
+  
+  if (showType) {
+    movieFromCountry = movieFromCountry.filter((item) => item.type ===  showType)
+  }
+
+  res.json(movieFromCountry)
+})
+
+// Query search on Titel always toLowerCase
+app.get('/search', (req, res) => {
```
idlefingers: I think this would be nice if it were in the `/movies` endpoint, too. As a rule of thumb, you only need to make new endpoints when it will return a different type of data, so a list of movies on `/movies`, a single movie on `/movies/:id`, movies for a country on `/movies?country=sweden`, or searching for a movie on `/movies?title=Jumanji`. If an endpoint will perform an action such as creating, updating or deleting an entity, it'll be a new endpoint, too.

PR #16, opened 2020-02-03
Pull request 

claragyokeres:


claragyokeres:
https://project-express-api-clara.herokuapp.com/

server.js
```diff
@@ -27,6 +15,23 @@ app.get('/', (req, res) => {
   res.send('Hello world')
 })
 
+app.get('/nominations', (req, res) => {
```
idlefingers: I get a server error when I try to load this endpoint :( https://project-express-api-clara.herokuapp.com/nominations

server.js
```diff
@@ -27,6 +15,23 @@ app.get('/', (req, res) => {
...
+  res.json(goldenGlobesDatadata)
+})
+
+app.get('/year/:year', (req, res) => {
```
idlefingers: Following the RESTful rules, I think this endpoint should be a filter on the `/nominations` endpoint rather than an endpoint of its own, since they both return nominations.

PR #17, opened 2020-02-03
AnnaTerelius project-express-api

AnnaTerelius:


JennieDalgren:
add link to deployed project please

AnnaTerelius:
https://tereliusexpressdeployment.herokuapp.com/

server.js
```diff
@@ -2,16 +2,12 @@ import express from 'express'
...
-// import avocadoSalesData from './data/avocado-sales.json'
-// import booksData from './data/books.json'
-// import netflixData from './data/netflix-titles.json'
-// import topMusicData from './data/top-music.json'
-
-// Defines the port the app will run on. Defaults to 8080, but can be 
+//deployed at: https://tereliusexpressdeployment.herokuapp.com/
+
+import netflixData from './data/netflix-titles.json'
+import { query } from 'express/lib/express'
```
JennieDalgren: remove this line `import { query } from 'express/lib/express'`

server.js
```diff
@@ -22,12 +18,75 @@ const app = express()
...
+ app.get('/show', (req, res) => { 
+ const titleSearchString = req.query.search;
+   let filteredShow = netflixData.filter(item => {
+     const itemTitle = item.title.toString();
+     return itemTitle.toLowerCase().includes(titleSearchString.toLowerCase())
+    
+   });
+  
+   res.json(filteredShow);
+ });
```
JennieDalgren: This is filtering the same data as you show in /films endpoint. So make this query be available in that endpoint instead of creating a new one. So you could either add i.e `?search=choc` to the /films endpoint or not. and it will return a filtered list or the whole list based on that.

server.js
```diff
@@ -22,12 +18,75 @@ const app = express()
...
+ //get the type of show with for example the path "type/Movie"
+app.get('/type/:type', (req, res) => {
+  const type = req.params.type
+  const typeOfFilm = netflixData.find((item) => item.type === type)
+  res.json(typeOfFilm)
+ })
+
+ //get the first 10 films with the path "films/firstTenFilms"
+ app.get('/films/:firstTenFilms', (req, res) => {
+  const firstTenFilms = netflixData.slice(0, 10)
```
JennieDalgren: nice that you're trying out the slice method!

server.js
```diff
@@ -22,12 +18,75 @@ const app = express()
...
+app.get('/film/:title', (req, res) => {
+  const title = req.params.title
+  const film = netflixData.filter((item) => item.title === title)
+  res.json(film)
+ })
+
+ //get the type of show with for example the path "type/Movie"
+app.get('/type/:type', (req, res) => {
+  const type = req.params.type
+  const typeOfFilm = netflixData.find((item) => item.type === type)
```
JennieDalgren: this will only return a single object, so if there are more shows with the same type, you'll only get the first match.
If you want to return multiple shows, use filter() method instead.

server.js
```diff
@@ -22,12 +18,75 @@ const app = express()
...
+  const searchString = req.query.search;
+    
+  let filteredShows = netflixData
+  
+  if (searchString) {
+    filteredShows = filteredShows.filter(item => {
+      const itemTitle = item.title.toString();
+      const itemCountry = item.country.toString();
+      const itemDescription = item.description.toString();
+      return itemTitle.toLowerCase().includes(searchString.toLowerCase()) || itemCountry.toLowerCase().includes(searchString.toLowerCase()) || itemDescription.toLowerCase().includes(searchString.toLowerCase())
```
JennieDalgren: ⭐️

PR #18, opened 2020-02-03
Project Express API - J.Fick

Jenfi:
Depoyed site  via heroku: https://jennifers-express-deploy.herokuapp.com/

Endpoints: 
- /shows
- /shows/:id
- /movies
- /moives/:id
- /year/:year

### DIFF
--- a/server.js
+++ b/server.js
@@ -1,10 +1,9 @@
 import express from 'express'
 import bodyParser from 'body-parser'
 import cors from 'cors'
+import netflixData from './data/netflix-titles.json'
 
-// If you're using one of our datasets, uncomment the appropriate import below
-// to get started!
-// 
+
 // {
 //   "show_id": 81193313,
 //   "title": "Chocolate",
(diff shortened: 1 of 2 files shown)


server.js
```diff
@@ -1,33 +1,48 @@
...
+  res.json(shows)
+})
+
+app.get('/shows/:id', (req, res) => {
+  const id = req.params.id
+  const showsId = netflixData.filter((shows) => shows.type === "TV Show" && shows.show_id === +id)
+  res.json(showsId)
+})
+
+app.get('/year/:year', (req, res) => {
```
idlefingers: I think, following REST, the `/shows`,  `/movies` and `/year/:year` endpoint should all be merged into one, and make use of query params to filter

* `/shows` - returns everything
* `/shows?type=tv` - returns only tv shows
* `/shows?type=movie` - returns only movies
* `/shows?year=2019` - returns anything from 2019

This way you could chain things together without having to duplicate lots of logic. Like 'show me movies from 2019' = `/shows?type=movie&year=2019`

PR #19, opened 2020-02-03
API using Express - Anna Almestål

anna-ip:
I had a hard time to grasp what the actual project was about or what it should include.
I've tried out a couple of things myself, didn't really get them to work and unfortunately I was very short of time last week, so didn't have time to experiment enough to be satisfied. 
But I have some gets and routes so I think I've understod the concept.

https://apinetflix.herokuapp.com/


server.js
```diff
@@ -1,33 +1,82 @@
...
+        itemDescrpition.toLowerCase().includes(searchString) ||
+        itemDirector.toLowerCase().includes(searchString);
+
+    })
+    res.json(filteredShows)
+  }
+})
+
+//filters with id
+app.get('/shows/id/:id', (req, res) => {
```
idlefingers: Following the REST rules, this should be `/shows/:id`

server.js
```diff
@@ -1,33 +1,82 @@
...
+  const id = req.params.id
+  // console.log(id)
+  const showId = netflixData.filter((item) => item.show_id === +id)
+  res.json(showId)
+})
+
+//filters with release year
+app.get('/shows/year/:year', (req, res) => {
```
idlefingers: I think it would be nice to merge this filter into the main `/shows` endpoint further up. So you could do `/shows?year=2019` and do the filtering in there.

PR #20, opened 2020-02-03
Project Express API, Michel Elmoznino Laufer

MichelLaufer:
Deployed project:
https://michel-book-api.herokuapp.com/books
Frontend-part:
https://michel-apiexpressproject.netlify.com/
https://github.com/MichelLaufer/project-api-express-books/tree/master/src

I think that this week's project was fun and interesting! I think that the structure of the project, video lectures and code-alongs were very good. It would be interesting to learn more about e.g. using post and delete methods for adding and deleting items from the API.


server.js
```diff
@@ -22,12 +13,60 @@ const app = express()
...
+app.get('/books1', (req, res) => {
+  console.log(books.slice(0, 251))
+  res.json(books.slice(0, 251))
+})
+
+// Return the second half of books (objects) from the array in books.json
+app.get('/books2', (req, res) => {
+  console.log(books.slice(251, 500))
+  res.json(books.slice(251, 500))
+})
```
idlefingers: It would be nice to see these endpoints tied into the `/books` endpoint, and use pagination to filter the number of results returned.

PR #21, opened 2020-02-03
Pull request Express API - Matilda Arvidsson.

HIPPIEKICK:
Such a fun week!

The only thing I'm not happy with is the duplicate code I have for the pagination, but I couldn't figure out how to make it work any other way. (Also the frontend was a lot just testing things out and playing around, so that is not complete in any way).

Didn't really know what to write instead of hello world on the home page, but here is heroku: [https://boardgames-matilda-arvidsson.herokuapp.com/](https://boardgames-matilda-arvidsson.herokuapp.com/)
here is the frontend on netlify: [https://matilda-arvidsson-boardgames.netlify.com/](https://matilda-arvidsson-boardgames.netlify.com/)
and here is the frontend repo if you wanna look: [https://github.com/HIPPIEKICK/Boardgames](https://github.com/HIPPIEKICK/Boardgames)

PS. Since the frontend was a lot of trying things out and playing around, I left it half finished, but maybe I will go back to it. For example the year and the rating params only work separately and not with pagination or name query. Stretch goal for the stretch goal haha.

### DIFF
--- a/server.js
+++ b/server.js
@@ -18,14 +18,12 @@ app.use(bodyParser.json())
 
 //Homepage
 app.get('/', (req, res) => {
-  res.send('Possible routes: /boardgames, /year/:year, /rating/:rating (change :rating to "average" or "rank" ')
+  res.send('Possible routes: /boardgames/:id, /year/:year, /rating/:rating (change :rating to "average" or "rank" ')
 })
 
 //Specific boardgames, requires id
 app.get("/boardgames/:id", (req, res) => {
   const id = req.params.id
-  // const headers = req.headers
-  // console.log("headers : " + JSON.stringify(headers, null, 2))
   const boardgame = boardgamesData.find((game) => game.id === +id)
   if (boardgame) {
     res.json(boardgame)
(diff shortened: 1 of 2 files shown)


server.js
```diff
@@ -22,9 +14,125 @@ const app = express()
...
+      res.json(filteredSearch.slice(pageSearch * 10 - 10, pageSearch * 10))
+    }
+  }
+
+  //Slice the data, begin on the page written in query
+  res.json(boardgamesData.slice(pageSearch * 10 - 10, pageSearch * 10))
+})
+
+//Boardgame from year, requires year
+app.get("/year/:year", (req, res) => {
```
idlefingers: It would be nice if this and the rating endpoints were all part of the `/boardgames` endpoint (and it'd be more RESTful).  One way to approach having multiple filters in one endpoint is to make a copy of the array of data and then slowly go through re-assigning it when there's some filter.  Something along these lines:

```js
let data = boardgamesData

if (nameSearch) {
  data = data.filter(...)
}

if (req.params.year) {
  data = data.filter(...)
}

if (req.params.rating) {
  data = data.filter(...)
}

res.json(data)
```

So you keep filtering and re-assigning the variable before finally returning it once all the possible filters have been run. Does that make sense?

PR #22, opened 2020-02-03
express-api-project

horsefacegirl:
https://horsefacegirl-express-deploy.herokuapp.com/

Didn't have so much time this week so just did the bare minimum. Will try to dig deeper next project.

### DIFF
--- a/server.js
+++ b/server.js
@@ -3,27 +3,14 @@ import bodyParser from "body-parser";
 import cors from "cors";
 import booksData from "./data/books.json";
 
-//     Example of json stucture
-//
-//     "bookID": 2,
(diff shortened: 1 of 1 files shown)


server.js
```diff
@@ -1,33 +1,37 @@
...
+  res.json(booksData);
+});
+
+app.get("/books/:id", (req, res) => {
+  const bookId = req.params.id;
+  const book = booksData.find(item => item.bookID.toString() === bookId);
+  res.json(book);
+});
+
+app.get("/language/:language", (req, res) => {
```
idlefingers: It would be nice if this was part of the `/books` endpoint (and it'd be more RESTful) - then make use of query params to filter on language. So you could have `/books` to return all books and then `/books?language=en` for english books, etc.

PR #23, opened 2020-02-03
Express API - Aleksandra

AleksandraMuller:
My deployment on Heroku: https://books-express-api.herokuapp.com/books
I created some query parameters and path parameters. Project was fun, when I will have some more time, I will add some frontend to it.

### DIFF
--- a/server.js
+++ b/server.js
@@ -2,15 +2,16 @@ import express from "express";
 import bodyParser from "body-parser";
 import cors from "cors";
 
+import booksData from "./data/books.json";
+
+// SAVING OTHER DATA FOR LATER DEVELOPMENT OF THE PROJECT
 // import goldenGlobesData from "./data/golden-globes.json";
 // import avocadoSalesData from './data/avocado-sales.json'
-import booksData from "./data/books.json";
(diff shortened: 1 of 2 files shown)


server.js
```diff
@@ -1,33 +1,93 @@
...
+
+  let info = booksData.filter(
+    item =>
+      item.average_rating === +searchRating || item.num_pages === +searchPages
+  );
+
+  res.json(info);
+});
+
+app.get("/books/pages/:pages", (req, res) => {
```
idlefingers: It would be more RESTful to have this filter as part of the `/books` endpoint and use query parameters to pass a pages param. Like `/books?pages=123` to filter where the number of pages is "123"

server.js
```diff
@@ -1,33 +1,93 @@
...
+
+app.get("/books/:id", (req, res) => {
+  const id = req.params.id;
+
+  let bookid = booksData.filter(item => item.bookID === +id);
+
+  res.json(bookid);
+});
+
+app.get("/books/authors/:author", (req, res) => {
```
idlefingers: I think this would also be well suited for a query param on the `/books` endpoint. Like `/books?author=Bob`

PR #24, opened 2020-02-03
Theresa Ulwahn - Project 17 Backend The Express API

TheresaUlwahn:
Hi there, 

struggled a little bit extra as I found an API of my own I wanted to make work. And made it.
Extra happy for that. :-) 

https://dashboard.heroku.com/apps/mrsucodes-express-api


server.js
```diff
@@ -1,33 +1,59 @@
...
+//Down below routes are defined
+// //http://localhost:8080/athletes
+// app.get('/athletes', (req, res) => {
+//   res.json(athletesData)
+// })
+
+app.get('/athletes/:id', (req, res) => {
+  const showId = req.params.id 
+  console.log(showId)
+  const show = athletesData.filter(item => +item.competitorid === +showId)
```

PR #25, opened 2020-02-03
Express api - Jenny Jacobsson

jennyjacobsson:
Hi!

My first backend project completed. Yay.

**My endpoints are:**
titles/ (all data)
titles/ search query for title, director or country
titles/:type (like TV Show or Movie)
id/:id

**Link to my project:**
https://projekt-16-express-api.herokuapp.com/

### DIFF
--- a/server.js
+++ b/server.js
@@ -24,7 +24,7 @@ app.use(bodyParser.json())
 
 // Start defining your routes here
 app.get('/', (req, res) => {
-  res.send('Hello world')
+  res.send('Possible path/routes to take: /titles, id/:id, titles/:type (like TV Show or Movie) and titles/ search query for title, director or country')
 })
 
 app.get('/titles', (req, res) => {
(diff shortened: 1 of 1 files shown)


server.js
```diff
@@ -24,10 +24,54 @@ app.use(bodyParser.json())
...
+        itemDirector.toLowerCase().includes(searchString) ||
+        itemCountry.toLowerCase().includes(searchString)
+    })
+  }
+  res.json(filtredTitles)
+})
+
+
+//Filter out type: TV Show or Movie (correct spelling needed)
+app.get('/titles/:type', (req, res) => {
```
idlefingers: It would be nice if this was part of the `/titles` endpoint (and it'd be more RESTful) - then make use of query params to filter on type. So you could have `/titles` to return all netflix titles and then `/titles?type=movie` for movies, etc. Then you'd be able to filter by both type and search for titles at the same time :)

server.js
```diff
@@ -24,10 +24,54 @@ app.use(bodyParser.json())
...
+  const type = req.params.type
+  let typeOf = netflixData.filter((item) => item.type === type)
+  res.json(typeOf)
+})
+
+
+//Find a specific title id (ex. 81193313 )
+app.get('/id/:id', (req, res) => {
```
idlefingers: To be RESTful, this endpoint should be `/titles/:id`

PR #26, opened 2020-02-03
Express API - Linda Isell

liissee:
Hej!
Here is my project :)
https://express-deployment-linda.herokuapp.com  
And here is a link to my front end code, didn't understand how to use/fetch the deployed herokuapp link. 
https://github.com/liissee/books/tree/master/src

I think I struggled the most with showing error messages when you filter on things,
Not related to this project, or a little maybe, but I'm still really interested in how to search and show things from an API using Redux.  :) 

All the best!
Linda

### DIFF
--- a/server.js
+++ b/server.js
@@ -7,7 +7,7 @@ import cors from 'cors'
 // 
 // import goldenGlobesData from './data/golden-globes.json'
 // import avocadoSalesData from './data/avocado-sales.json'
-// import booksData from './data/books.json'
+import booksData from './data/books.json'
 // import netflixData from './data/netflix-titles.json'
 // import topMusicData from './data/top-music.json'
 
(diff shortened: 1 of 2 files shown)


server.js
```diff
@@ -1,30 +1,111 @@
...
+    //if array is empty show error message
+  } if (filteredBooks.length === 0) {
+    res.status(404).send(`No books found`);
+  } else {
+    res.json(filteredBooks)
+  }
+})
+
+//Filters on language_code.  http://localhost:8080/language/eng
+app.get('/language/:language', (req, res) => {
```
idlefingers: It would be nice if this was part of the `/books` endpoint (and it'd be more RESTful) - then make use of query params to filter on language. So you could have `/books` to return all books and then `/books?language=en` for english books, etc.

server.js
```diff
@@ -1,30 +1,111 @@
...
+  if (bookLanguage.length === 0) {
+    res.status(404).send(`No books with the language: ${language} found.`);
+
+  } else {
+    res.json(bookLanguage)
+  }
+})
+
+//Sort books from highest rating to lowest. http://localhost:8080/sort/rating
+app.get('/rating', (req, res) => {
```
idlefingers: I think it'd be nice if this was also a filter on the `/books` endpoint. Like `/books?rating=5`

server.js
```diff
@@ -1,30 +1,111 @@
...
+    text: "Sorted by average rating",
+    booksByRating
+  })
+})
+
+
+//PAGINATION. Show 20 books per page. URL = http://localhost:8080/books1?page=1
+const PER_PAGE = 20
+
+app.get('/page', (req, res) => {
```
idlefingers: This too - you could keep using query parameters to avoid needing new routes which return the same type of data. As a rule of thumb, your routes should be named after the data they expect to return, rather than how they're going to sort/filter the data. If the endpoint is going to return an array of books, it should be the `/books` endpoint. If it's gonna return a single book, then it should be part of the `/books/:id` endpoint, etc.

PR #27, opened 2020-02-03
Nina's Book API

ninamansster:
Here comes my Book API in Express Server.
Check it out here:
https://ninamansster-book-api.herokuapp.com/

### DIFF
--- a/server.js
+++ b/server.js
@@ -24,7 +24,9 @@ app.use(bodyParser.json())
 
 // Start defining your routes here
 app.get('/', (req, res) => {
-  res.send('Hello planet Earth!')
+  // res.send('Hello planet Earth!')
+  res.send("Try searches based on the routes: /books, /id/[a number], /author/[name of author], /titles/[name of title] or  query for any word in title or author: library?search=[type your search]")
+
 })
 
 // Example of JSON-structure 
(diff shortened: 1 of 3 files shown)


server.js
```diff
@@ -24,9 +24,114 @@ app.use(bodyParser.json())
...
+})
+
+
+// PAGE - to have the result on a page with x number of boooks.
+
+const PER_PAGE = 10
+
+
+// routte to GET books by page (http://localhost:8080/booklists?page=2)
+app.get('/booklists', (req, res) => {
```
idlefingers: Don't forget about REST. So this endpoint should be `/books`

server.js
```diff
@@ -24,9 +24,114 @@ app.use(bodyParser.json())
...
+
+  res.json({
+    totalPages: Math.floor(booksData.length / PER_PAGE),
+    currentPage: +page,
+    data
+  })
 })
 
+//creating a sorted rating (localhost:8080/books/sort/rating)
+app.get('/books/sort/rating', (req, res) => {
```
idlefingers: It would be nice if this was part of the `/books` endpoint (and it'd be more RESTful) - then make use of query params to filter on rating. So you could have `/books` to return all books and then `/books?rating=5` etc

server.js
```diff
@@ -24,9 +24,114 @@ app.use(bodyParser.json())
...
+  const booksByRating = booksData.sort((a, b) => -(parseFloat(a.average_rating) - parseFloat(b.average_rating)))
+  res.json(booksByRating.slice(0, 50))
+})
+
+
+//creating a rest route for id (localhost:8080/books/5) https://sv.wikipedia.org/wiki/404_error
+app.get('/books/id/:id', (req, res) => {
```
idlefingers: This endpoint should just be `/books/:id`

server.js
```diff
@@ -24,9 +24,114 @@ app.use(bodyParser.json())
...
+  const findThatBook = booksData.find((item) => item.bookID === +bookId)
+  if (findThatBook) {
+    res.json(findThatBook)
+  } else {
+    res.status(404).send('There was no book with that id')
+  }
+})
+
+//creating a rest route for id (localhost:8080/titles/cc), https://sv.wikipedia.org/wiki/404_error
+app.get('/titles/:titles', (req, res) => {
```
idlefingers: This should also be a query param on the `/books` endpoint. Like `/books?title=hello`

server.js
```diff
@@ -24,9 +24,114 @@ app.use(bodyParser.json())
...
+  const findThatBookTitle = booksData.find((book) => book.title === titles)
+  if (findThatBookTitle) {
+    res.json(findThatBookTitle)
+  } else {
+    res.status(404).send('There was no book with that title, try to search with /library?search=[your search] instead. ')
+  }
+})
+
+//creating a rest route called author localhost:8080/author/Exact name of author
+app.get('/author/:author', (req, res) => {
```
idlefingers: This too! :) `/books?author=Bob`

PR #28, opened 2020-02-03
Avocado ftw

johannawaern:


server.js
```diff
@@ -23,8 +15,59 @@ app.use(cors())
...
+      (sale) => sale.totalBagsSold >= totalBagsSoldMoreThan
+    )
+    res.json(matchingSales)
+  } else {
+    res.json(sales)
+  }
+})
+
+// average price, sorted from low to high
+app.get('/salesByAveragePrice', (req, res) => {
```
idlefingers: It would be nice if this was part of the `/sales` endpoint (and it'd be more RESTful) - then make use of query params to filter on price. So you could have `/sales` to return all sales and then `/sales?averagePrice=12` for filtering on price

server.js
```diff
@@ -23,8 +15,59 @@ app.use(cors())
...
+
+// /sales/5 - sale with id 5 in every region
+app.get('/sales/:id', (req, res) => {
+  const id = req.params.id
+  const saleIdInAllRegions = sales.filter((sale) => sale.id === +id)
+  res.json(saleIdInAllRegions)
+})
+
+// /Albany/sales - all sales in Albany
+app.get('/:region/sales', (req, res) => {
```
idlefingers: Nice! This could also be a filter on the `/sales` endpoint but having the region in front makes it still feel RESTful so I'm not going to complain ;)

PR #29, opened 2020-02-03
Express-api-netflix 

petrabrinck:
Fun project but we took a little detour and fetched our own api:s. In the end it turned out to to be a great lesson ;-)

### DIFF
--- a/server.js
+++ b/server.js
@@ -1,7 +1,9 @@
 import express from 'express'
 import bodyParser from 'body-parser'
 import cors from 'cors'
+import netflixData from './data/netflix-titles.json'
 
+console.log(netflixData.length)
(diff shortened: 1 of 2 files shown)


server.js
```diff
@@ -24,8 +19,105 @@ app.use(bodyParser.json())
...
+
+//http://localhost:8080/country/Sweden
+// app.get('/country/:country', (req, res) => {
+//   const country = req.params.country
+//   const fromCountry = netflixData.filter((item) => item.country === country)
+//   res.json(fromCountry)
+// })
+
+//Filter based on title OR Country or decription
+app.get('/shows', (req, res) => {
```
idlefingers: This is using the same path for the endpoint, so I think it'd be much clearer to have this filtering inside the first `/shows` endpoint.

PR #30, opened 2020-02-04
Frida Project Express API

annafridamaria:
[https://frijon-art-api.herokuapp.com/](https://frijon-art-api.herokuapp.com/)

Felt kind of scary to dip my toes into backend, but I will definitely go back to the friday codealong to try and link it to a front end

Can one use two methods on one object? Because now the search is case sensitive, and i would like to make both the search string and the possible matches into lowercase, but it felt more relevant to make sure they were actually strings (ok, in this case it was names, genres or countries but if it was a music api and Prince would still be a symbol....)

### DIFF
--- a/server.js
+++ b/server.js
@@ -19,15 +19,7 @@ app.get('/', (req, res) => {
   res.send('Hello world')
 })
 
-// app.get('/art', (req, res) => {
-//   // res.send('Hello world')
-//   res.json(data)
-//   // genre
-//   // country
-// })
-
 app.get('/genre/:genre', (req, res) => {
-// artist name
 const genre = req.params.genre
 const showGenre = data.filter(artist => artist.genre === genre)
 res.json(showGenre)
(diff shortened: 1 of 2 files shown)


server.js
```diff
@@ -27,6 +19,36 @@ app.get('/', (req, res) => {
   res.send('Hello world')
 })
 
+app.get('/genre/:genre', (req, res) => {
```
idlefingers: I think it would be nice if this was part of the `/art` endpoint (and it'd be more RESTful) - then make use of query params to filter on genre. So you could have `/art` to return all art pieces and then `/art?genre= Surrealism`, etc.

In the RESTful world, we use the endpoints to describe the data returned, so for a `/genre` endpoint, you'd expect information about the genre, not pieces of art. if the endpoint was `/genres/:genre/art` then you'd expect art peices since the last part is 'art'. Does that make sense?

server.js
```diff
@@ -27,6 +19,36 @@ app.get('/', (req, res) => {
...
+const genre = req.params.genre
+const showGenre = data.filter(artist => artist.genre === genre)
+res.json(showGenre)
```
idlefingers: Keep an eye on your indentation

PR #31, opened 2020-02-05
Express-API - Disa

Disa-san:


server.js
```diff
@@ -27,6 +30,72 @@ app.get('/', (req, res) => {
...
+//   "average_rating": 4.33,
+//   "isbn": 374518734,
+//   "isbn13": 9780374518738,
+//   "language_code": "eng",
+//   "num_pages": 720,
+//   "ratings_count": 2966,
+//   "text_reviews_count": 211
+// },
+
+app.get('/allBooks', (req, res) => {
```
JennieDalgren: call this endpoint `/books` instead of allBooks

server.js
```diff
@@ -27,6 +30,72 @@ app.get('/', (req, res) => {
...
+
+// path variable/path parameter
+// app.get('/book/:id', (req, res) => {
+//   const bookId = req.params.id
+//   const book = booksData.filter(item => item.bookID === +bookId)
+//   res.json(book) §§
+// })
+
+// query parameter
+app.get('/author', (req, res) => {
```
JennieDalgren: Since this is filtering from the booksData, make these queries in the books endpoint instead of creating a new one.

PR #32, opened 2020-02-06
Books API

JoacimNilsen:
Here is my project!
https://expressapi-books.herokuapp.com/

I had some problems with the syntax. Especially for the `/title/:title` where i wanted to make it so that you could go to books starting with a certain letter. I knew i wanted to use charAt, but i kept getting errors.
I asked Tiago in my group for help and we figured it out together!

I also wanted to do paging but i did not have enough time. I did the code along from friday and kept getting empty arrays back. Dropped that for now but I am still very satisfied with the result.


server.js
```diff
@@ -1,33 +1,51 @@
...
+  const language = req.params.language
+  const booksByLanguage = booksData.filter(item => item.language_code === language)
+  res.json(booksByLanguage)
+})
+
+app.get('/sort/rating', (req, res) => {
+  const copyBooksData = [...booksData]
+  const sortByRating = copyBooksData.sort((a, b) => -(parseFloat(a.average_rating)-parseFloat(b.average_rating)))
+  res.json(sortByRating)
+})
```
JennieDalgren: books/rating might be a better name for this route

server.js
```diff
@@ -1,33 +1,51 @@
...
+
+app.get("/title/:title", (req, res) => {
+  const title = req.params.title
+  const titleLetter = booksData.filter(book => {
+    let bookTitle = book.title.toString()
```
JennieDalgren: same as for language, add this as a query of `/books` instead.

server.js
```diff
@@ -1,33 +1,51 @@
...
+  const bookId = req.params.id
+  const book = booksData.find(item => item.bookID.toString() === bookId)
+  res.json(book)
+})
+
+app.get("/language/:language", (req, res) => {
+  const language = req.params.language
+  const booksByLanguage = booksData.filter(item => item.language_code === language)
+  res.json(booksByLanguage)
+})
```
JennieDalgren: add this as a query to the /books endpoint instead. To either get all books, or if you add a query such as `books?language=swedish` you filter out them and return the filtered books instead.

Same with title further down.

PR #36, opened 2020-02-17
Late PR - Express API - Golden Globes

ucha3i:
I'm pretty sure, I messed up something with pull request here, but don't really know why... 

Did I fork the wrong repo?

### DIFF
--- a/server.js
+++ b/server.js
@@ -1,74 +1,39 @@
 import express from 'express'
 import bodyParser from 'body-parser'
 import cors from 'cors'
-import sales from './data/avocado-sales.json'
+/* import goldenGlobesData from './data/golden-globes.json' */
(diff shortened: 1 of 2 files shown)


server.js
```diff
@@ -1,32 +1,51 @@
...
+
+  if (titleSearchString) {
+    filteredShows = filteredShows.filter(item => {
+      const itemTitle = item.title.toString();
+        return itemTitle.includes(titleSearchString)
+    });
+  }
+
+  res.json(filteredShows);
+  }); */
```
JennieDalgren: why is this commented out?

ucha3i: Because I didn't want to work with netflix data, but  also didn't want to remove it, because it works and I can use it later for other projects.

PR #41, opened 2020-05-04
Express api books Hanna Rüter (review wanted)

Ruterryter:
Hi, 
Here is a link to this weeks project. 

https://express-book-api-hannaruter.herokuapp.com/

I really enjoyed trying out some backend and looking forward to more.  I really struggled with the query and tested a million things before I got it to work as I wanted. But with some help from my team it worked at last.

### DIFF
--- a/server.js
+++ b/server.js
@@ -25,12 +25,26 @@ app.use(bodyParser.json())
 
 //Start defining your routes here
 app.get('/', (req, res) => {
-  res.send('Endpoints:  /books, /books/:id, /ratings, authors/:author')
+  res.send('Endpoints:  /books, /books/book, /books/:id, /ratings, authors/:author')
 })
 
 //List all the books in the api
 app.get('/books', (req, res) => {
   res.json(booksData)
+})
+
+//Query for title (can only find titles with lowercase and 1 word)
+app.get('/books/book', (req, res) => {
+  const title = req.query.title
+  let bookTitle = booksData.find((item) =>
+    item.title.toString().toLowerCase().includes(title))
+
+  if (bookTitle) {
+    res.json(bookTitle)
+
+  } else {
+    res.status(404).json({ message: 'Not found' })
+  }
 })
 
 //Show one book using an :id as a placeholder for the bookID number in the data
@@ -44,14 +58,15 @@ app.get('/books/:id', (req, res) => {
   }
 })
 
-//Sort books after rating, highest to lowest
+
+//Sort books after rating, highest to lowest (make this into a query in book)
 app.get('/ratings', (req, res) => {
   const bestRated = booksData.sort((a, b) => b.average_rating - a.average_rating);
   res.json(bestRated)
 })
 
 
-//Filter books by author
+//Filter books by one author
 app.get('/authors/:author', (req, res) => {
   const author = req.params.author;
   const booksByAuthor = booksData.filter((item) =>
@@ -59,9 +74,6 @@ app.get('/authors/:author', (req, res) => {
   );
   res.json(booksByAuthor);
 });
-
-
-//Query by title
 
 
 // Start the server
--- a/server.js
+++ b/server.js
@@ -25,12 +25,24 @@ app.use(bodyParser.json())
 
 //Start defining your routes here
 app.get('/', (req, res) => {
-  res.send('Endpoints:  /books, /titles, /books/:id, /ratings, authors/:author')
+  res.send('Endpoints:  /books /books/:id, /ratings, authors/:author')
 })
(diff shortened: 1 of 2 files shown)


server.js
```diff
@@ -2,31 +2,81 @@ import express from 'express'
...
 
 // Add middlewares to enable cors and json body parsing
 app.use(cors())
 app.use(bodyParser.json())
 
-// Start defining your routes here
+//Start defining your routes here
 app.get('/', (req, res) => {
-  res.send('Hello world')
+  res.send('Endpoints:  /books, /books/:id, /ratings, authors/:author')
```
puj, line 28: Love this idea! Would be nice with a bit of formatting

server.js
```diff
@@ -2,31 +2,81 @@ import express from 'express'
...
+//Sort books after rating, highest to lowest 
+app.get('/ratings', (req, res) => {
+  const bestRated = booksData.sort((a, b) => b.average_rating - a.average_rating);
+  res.json(bestRated)
+})
+
+
+//Get all books by one author (example: https://express-book-api-hannaruter.herokuapp.com/authors/rowling)
+app.get('/authors/:author', (req, res) => {
+  const author = req.params.author;
```
puj, line 72: When pointing to a specific resource in the endpoint, such as an author, it's really good to use a path parameter like you have here.

Your `author` path parameter filters a result instead of "identifying" a result.

For instance 
`GET /authors/dan` returns books from 
Dan Brown and Daniel Abraham.  

A **path parameter** is a good option if you are using it to point to a specific author, however **query parameters** are more optimal when filtering results which could return more than one different author.

server.js
```diff
@@ -2,31 +2,81 @@ import express from 'express'
...
+  const booksByAuthor = booksData.filter((item) =>
+    item.authors.toLowerCase().includes(author)
```
puj, line 74: I like this idea.  What if:
`author = "Adam"`

Will this catch a book with the author "Adam"?

server.js
```diff
@@ -2,31 +2,81 @@ import express from 'express'
...
+  );
+  res.json(booksByAuthor);
```
puj, line 76: `booksByAuthor ` is a VERY strong name for this variable.

server.js
```diff
@@ -2,31 +2,81 @@ import express from 'express'
...
+import booksData from './data/books.json'
+//import netflixData from './data/netflix-titles.json'
 // import topMusicData from './data/top-music.json'
 
 // Defines the port the app will run on. Defaults to 8080, but can be 
 // overridden when starting the server. For example:
 //
 //   PORT=9000 npm start
-const port = process.env.PORT || 8080
+const port = process.env.PORT || 8088
```
puj, line 19: Why 8088? Favorite number? 😄

server.js
```diff
@@ -2,31 +2,81 @@ import express from 'express'
...
-  res.send('Hello world')
+  res.send('Endpoints:  /books, /books/:id, /ratings, authors/:author')
 })
 
+//List all the books in the api with /books and query for title
+
+app.get('/books', (req, res) => {
+  const { title } = req.query
+  const bookTitle = booksData.filter((item) =>
+    item.title.toString().toLowerCase().includes(title))
```
puj, line 36: Careful with capital letters in the title query parameter 🔍

server.js
```diff
@@ -2,31 +2,81 @@ import express from 'express'
...
 app.get('/', (req, res) => {
-  res.send('Hello world')
+  res.send('Endpoints:  /books, /books/:id, /ratings, authors/:author')
 })
 
+//List all the books in the api with /books and query for title
+
+app.get('/books', (req, res) => {
+  const { title } = req.query
+  const bookTitle = booksData.filter((item) =>
```
puj, line 35: `bookTitle` makes me think this variable is a String

`books`
`filteredBooks`
 are some name suggestions

server.js
```diff
@@ -2,31 +2,81 @@ import express from 'express'
...
+    res.status(404).json({ message: 'Not found' })
+  }
+})
+
+
+//Sort books after rating, highest to lowest 
+app.get('/ratings', (req, res) => {
+  const bestRated = booksData.sort((a, b) => b.average_rating - a.average_rating);
+  res.json(bestRated)
+})
```
puj, line 67: Really cool! Well done.

server.js
Ruterryter: True, I will change that.

server.js
Ruterryter: I'm not really sure what you mean by this, where is the problem with the capital letters here?

PR #42, opened 2020-05-04
Express API - Christina Persson: Review wanted

Cilpersson:
The only thing that was unclear during this project was why I needed to use toString on something that most definitely looked like a string. I added a comment in the code about it.  

EDIT: Learned today that a few titles are numbers and not strings, that's why toString is needed! 

https://perssons-express-api.herokuapp.com/

### DIFF
--- a/server.js
+++ b/server.js
@@ -16,23 +16,13 @@ const app = express();
 app.use(cors());
 app.use(bodyParser.json());
 
+const listEndpoints = require("express-list-endpoints");
+
 app.get("/", (req, res) => {
-  // Displays all movies and shows
-  res.json(data);
+  res.send(listEndpoints(app));
 });
 
-// Displays object(s) with specific title
-app.get("/titles/:title", (req, res) => {
-  const title = req.params.title;
-  // I dont know why I need to use toString on item.title and not on item.actore further down, would love to know why!
-  const contentWithTitle = data.filter(
-    (item) => item.title.toString().toLowerCase() === title.toLowerCase()
-  );
-
-  res.json(contentWithTitle);
-});
-
-app.get("/IDs/:id", (req, res) => {
+app.get("/content/:id", (req, res) => {
   // Displays object with specific ID
   const id = req.params.id;
   const contentWithId = data.filter((item) => item.show_id === +id);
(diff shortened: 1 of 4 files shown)


server.js
```diff
@@ -1,33 +1,94 @@
...
+});
+
+app.get("/content", (req, res) => {
+  // Sorts based on queries
+  const { page, actor, type, title, year, country, director } = req.query;
+  let dataList = data;
+
+  if (actor) {
+    dataList = dataList.filter((item) =>
+      item.cast.toLowerCase().includes(actor.toLowerCase())
```
puj, line 32: What if you had a function 
`containsIgnoreCase(item.cast, actor)`?

server.js
```diff
@@ -1,33 +1,94 @@
...
+    dataList = dataList.filter((item) =>
+      item.country.toString().toLowerCase().includes(country.toLowerCase())
+    );
+  }
+
+  if (director) {
+    dataList = dataList.filter((item) =>
+      item.director.toString().toLowerCase().includes(director.toLowerCase())
+    );
+  }
```
puj, line 62: This code is so clear.  Naming is simple and effective. 
The structure is easy on the eyes.  See my comment about a reusable function above ☝️

server.js
```diff
@@ -1,33 +1,94 @@
...
+
+  const startIndex = 10 * (+page - 1) || 0;
+  const endIndex = startIndex + 10;
```
puj, line 65: `10` should be its own variable.  It will make the code more readable and prevent literal duplication

`(+page - 1)` warrants a comment.  What if the programmer reading this has never done their own paging before?

I'm not sure what the `|| 0` is doing here?  Ignoring the value of `page` if it's not valid input?

server.js
```diff
@@ -1,33 +1,94 @@
...
+  const dataListPaginated = dataList.slice(startIndex, endIndex);
```
puj, line 66: Fantastic naming.

server.js
```diff
@@ -1,33 +1,94 @@
...
+
+  if (dataListPaginated.length === 0) {
+    res.status(404).json({
+      error: "No shows or series found, please try a different query",
```
puj, line 70: Be careful with string literals.  In fact, be careful of all literals.

Move towards the idea that everything should be a constant or a variable.
```javascript
res.status(ERROR_CODE_NOT_FOUND).json({
  error: ERROR_MESSAGE_NO_SHOWS_FOUND
})
```
☝️ This might be common in the industry

Why?  Perhaps you need to introduce multiple languages..
```javascript
res.status(ERROR_CODE_NOT_FOUND).json({
  // Look up the correct translation based on `local` and `KEY`
  error: translate(ERROR_MESSAGE_NO_SHOWS_FOUND_KEY)
})
```

server.js
```diff
@@ -1,33 +1,94 @@
...
+    });
+  }
+
+  res.json({ content: dataListPaginated });
```
puj, line 74: `content` 
`data`
`results`
`result`

All good

server.js
```diff
@@ -1,33 +1,94 @@
...
+});
+
+app.get("/content/:id", (req, res) => {
+  // Displays object with specific ID
+  const id = req.params.id;
+  const contentWithId = data.find((item) => item.show_id === +id);
```
puj, line 80: `contentById`
`contentByID`

Minor

server.js
Cilpersson: I've done this in a much better way this week, and by doing so I understand it much better!

server.js
Cilpersson: Oh wow! I never even thought of that 🧠💥

PR #43, opened 2020-05-05
Frida Express API

fridamaria:
https://fridamaria-book-api.herokuapp.com/

This week's project was fun, I enjoyed testing some backend dev. I first tried out more routes but decided to include them as queries to the /books route instead. It would have been fun to test out some other methods as well, but one step at a time.

I look forward to learning more!

### DIFF
--- a/server.js
+++ b/server.js
@@ -14,12 +14,13 @@ const app = express()
 app.use(cors())
 app.use(bodyParser.json())
 
+const listEndpoints = require('express-list-endpoints')
+
 // Start defining your routes here
 
 // Root endpoint
 app.get('/', (req, res) => {
-  res.send(
-    'Available book API endpoints: /books, /books/:id, /min_pages/:pages, /min_rating/:rating')
+  res.send(listEndpoints(app))
 })
 
 // Route for array of all books with pagination as default
@@ -46,7 +47,7 @@ app.get('/books/:id', (req, res) => {
 })
 
 // Route for books based on min_pages. Accepting filter via query for max_pages, returning array of books between min and max no of pages
-app.get('/min_pages/:pages', (req, res) => {
+app.get('/books/min_pages/:pages', (req, res) => {
   const minPages = req.params.pages
   const maxPages = req.query.max_pages
   let numPages = booksData.filter((item) => item.num_pages >= +minPages)
@@ -59,7 +60,7 @@ app.get('/min_pages/:pages', (req, res) => {
 })
 
 // Same concept as route for pages but for rating
-app.get('/min_rating/:rating', (req, res) => {
+app.get('/books/min_rating/:rating', (req, res) => {
   const minRating = req.params.rating
   const maxRating = req.query.max_rating
   let bookRating = booksData.filter((item) => item.average_rating >= +minRating)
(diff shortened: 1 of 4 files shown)


server.js
```diff
@@ -1,30 +1,74 @@
...
-//   PORT=9000 npm start
+// PORT=9000 npm start
 const port = process.env.PORT || 8080
 const app = express()
 
 // Add middlewares to enable cors and json body parsing
 app.use(cors())
 app.use(bodyParser.json())
 
+const listEndpoints = require('express-list-endpoints')
```
puj, line 17: Nice! This is super cool!

server.js
```diff
@@ -1,30 +1,74 @@
...
+app.get('/books', (req, res) => {
+  const { page, sort, author, title } = req.query
+  let booksList = booksData
+
+  // sort books by rating using query, returns sorted booksList
+  if (sort === 'rating_dsc') {
+    booksList = booksList.sort((a, b) => (b.average_rating - a.average_rating))
+  } else if (sort === 'rating_asc') {
+    booksList = booksList.sort((a, b) => (a.average_rating - b.average_rating))
+  }
```
puj, line 36: Because you have a value here called `rating_asc`, I feel like there might be a `title_desc` in the future.  Was that the idea with the naming?

server.js
```diff
@@ -1,30 +1,74 @@
...
+
+  // find books by author using query, returns filtered booksList
+  if (author) {
+    booksList = booksList.filter((item) => item.authors.toString().toLowerCase().includes(author.toLowerCase()))
+  }
```
puj, line 41: Very nice.

server.js
```diff
@@ -1,30 +1,74 @@
...
+  const startIndex = 10 * (+page - 1) || 0
+  const endIndex = startIndex + 10
+  const booksListPaginated = booksList.slice(startIndex, endIndex)
+
+  // 404 when there are no books in the list
+  if (booksListPaginated.length === 0) {
+    res.status(404).json({ error: 'No books found, please try a different query' })
+  }
+
+  res.json({ books: booksListPaginated })
```
puj, line 58: I like the choice of naming this `books`

server.js
```diff
@@ -1,30 +1,74 @@
...
+// Route for single book, filtered using book id
+app.get('/books/:id', (req, res) => {
+  const { id } = req.params
+  const book = booksData.find((item) => item.bookID === +id)
+
+  // 404 when there is no book with path id
+  if (!book) {
+    res.status(404).json({ error: 'There is no book with that id' })
+  }
+
```
puj, line 70: Simple and effective!

server.js
```diff
@@ -1,30 +1,74 @@
...
+  // find books based on words in title using query, returns filtered booksList
+  if (title) {
+    booksList = booksList.filter((item) => item.title.toString().toLowerCase().includes(title.toLowerCase()))
+  }
+
+  // slice used for pagination
+  const startIndex = 10 * (+page - 1) || 0
+  const endIndex = startIndex + 10
+  const booksListPaginated = booksList.slice(startIndex, endIndex)
```
puj, line 51: The `startIndex` could use a comment explaining the math going on there 👍

server.js
fridamaria: Had an idea of doing more, ended up with just this one though. Added some more now!

server.js
fridamaria: Added some comments 😊

PR #46, opened 2020-05-06
Review wanted: week17 - Project Express API - Bea

bealun:
I struggle mostly with time management this week and really had to push through. This will be a project where I'll go back and clean up - especially the frontend. 
I'm really excited about backend and this was a great start. It's highly logical but also tricky since it's kind of abstract.

I couldn't get an error to work if a user tried a non-existing id. The code is cleaned up now but I've tried a lot of length-methods or to throw an error message if the id isn't between 1-50 so I would like to accomplish that with a bit of help.

View it live:

Frontend: https://music-express-bealun.netlify.app/ 
Backend: https://bealun-express-api.herokuapp.com/

This is the preliminary frontend to look at where I've used this API: https://github.com/bealun/music-express-frontend

### DIFF
--- a/server.js
+++ b/server.js
@@ -33,18 +33,30 @@ app.get('/', (req, res) => {
 
 
 // Show tops songs 
-// Can filter on genres included in query
+// Can filter on queries
 app.get('/song', (req, res) => {
   const genre = req.query.genre
+  const popularity = req.query.popularity
 
   let filteringMusic = topMusicData
 
+  // Filter on genre
   if (genre) {
     filteringMusic = filteringMusic.filter(song => {
       const genreTitle = song.genre
       return genreTitle.includes(genre)
     }
   )}
+
+  // Filter on popularity, high or low
+  if (popularity) {
+    if (popularity === 'high') {
+      filteringMusic.sort((a, b) => b.popularity - a.popularity)
+    } else if (popularity === 'low') {
+      filteringMusic.sort((a, b) => a.popularity - b.popularity)
+    }
+  } 
+
   res.json(filteringMusic)
 })
 
--- a/server.js
+++ b/server.js
@@ -1,15 +1,19 @@
 import express from 'express'
 import bodyParser from 'body-parser'
 import cors from 'cors'
+// import goldenGlobesData from './data/golden-globes.json'
+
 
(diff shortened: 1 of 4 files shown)


server.js
```diff
@@ -24,9 +16,50 @@ app.use(bodyParser.json())
...
 // Start defining your routes here
 app.get('/', (req, res) => {
-  res.send('Hello world')
+  res.send('Hello world, my first backend!')
+})
+
+
+// Show tops songs 
+// Can filter on queries
+app.get('/song', (req, res) => {
```
HIPPIEKICK, line 25: Pluralize the routes: `/songs`

server.js
```diff
@@ -24,9 +16,50 @@ app.use(bodyParser.json())
...
+  )}
+
+  // Filter on popularity, high or low
+  if (popularity) {
+    if (popularity === 'high') {
+      filteringMusic.sort((a, b) => b.popularity - a.popularity)
+    } else if (popularity === 'low') {
+      filteringMusic.sort((a, b) => a.popularity - b.popularity)
+    }
+  }
```
HIPPIEKICK, line 46: Supe nice with multiple queries! ⭐

server.js
```diff
@@ -24,9 +16,50 @@ app.use(bodyParser.json())
...
 
+// Shows specific song with info
+app.get('/song/:id', (req, res) => {
+  const id = req.params.id
+
+  let showSongs = topMusicData.filter((song) =>
+    song.id === +id)
+
+  res.json(showSongs)
+})
```
HIPPIEKICK, line 59: Nice! Using `filter()`, you will get a filtered array as a return. If you instead use `find()`, you would get a single value from the array, and you should be able to use that to check if the id (song) exists or not:

```
app.get("/songs/:id", (req, res) => {
  const id = req.params.id
  let showSongs = topMusicData.find((song) => song.id === +id)
  if (showSongs) {
    res.json(showSongs)
  } else {
    res.status(404).send(`No song found with id: ${id}`)
  }
})
```

HIPPIEKICK: The variable showSongs could be more descriptive of what it contains - it just returns a single song, so why not call it song 😎

HIPPIEKICK:
Also - super nice that you did a frontend!!

PR #47, opened 2020-05-06
Sara (review wanted)

Aqveduktis:
Live:
https://express-books-aqv.herokuapp.com/

I'm happy with how it turned out, could probably do more but this works. 

### DIFF
--- a/server.js
+++ b/server.js
@@ -36,12 +36,12 @@ app.get('/books', (req, res) => {
 
 app.get('/books/id/:id', (req, res) => {
   const id = req.params.id
-  const bookId = data.find((item)=>item.bookID == id)
-  if (bookId) {
-  res.json(bookId)
+  const book = data.find((item)=>item.bookID == id)
+  if (book) {
+  res.json(book)
   }
   else {
-    res.send("404 not found")
+    res.status(404).json({error:"book not found"})
   }
 })
 
@@ -52,7 +52,7 @@ app.get('/books/authors/:author', (req, res) => {
     res.json(booksAuthor)
   }
   else {
-    res.send('404 not found')
+    res.status(404).json({error:"book not found"})
   }
   
 })
--- a/server.js
+++ b/server.js
@@ -1,6 +1,8 @@
 import express from 'express'
 import bodyParser from 'body-parser'
 import cors from 'cors'
+import data from './data/books.json'
+
 
 // If you're using one of our datasets, uncomment the appropriate import below
 // to get started!
@@ -18,6 +20,7 @@ import cors from 'cors'
 const port = process.env.PORT || 8080
 const app = express()
 
+
 // Add middlewares to enable cors and json body parsing
 app.use(cors())
 app.use(bodyParser.json())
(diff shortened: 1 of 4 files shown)


server.js
```diff
@@ -1,30 +1,60 @@
...
 
+
 // Add middlewares to enable cors and json body parsing
 app.use(cors())
 app.use(bodyParser.json())
 
 // Start defining your routes here
 app.get('/', (req, res) => {
-  res.send('Hello world')
+  res.send('Try /books')
```
HIPPIEKICK, line 17: If you want to document it even more:
[https://www.npmjs.com/package/express-list-endpoints](https://www.npmjs.com/package/express-list-endpoints)

server.js
```diff
@@ -1,30 +1,60 @@
...
+  page = (page<=maxPage && page>=1)?page : 1;
+  const startIndex = (page -1) * pageSize 
+  const endIndex = startIndex + pageSize
+  const resultPage = data.slice(startIndex, endIndex)
+  const result = {numBooks: resultPage.length, 
+  page : page,
+  maxPage: maxPage,
+  books: resultPage}
+  res.json(result)
+})
```
HIPPIEKICK, line 35: Super nice with pagination!

server.js
```diff
@@ -1,30 +1,60 @@
...
+
+app.get('/books/id/:id', (req, res) => {
```
HIPPIEKICK, line 37: You could just write like this:
`/books/:id`

server.js
```diff
@@ -1,30 +1,60 @@
...
+  const id = req.params.id
+  const bookId = data.find((item)=>item.bookID == id)
+  if (bookId) {
+  res.json(bookId)
+  }
+  else {
+    res.send("404 not found")
```
HIPPIEKICK, line 44: You could send a status code and a message like this:
`res.status(404).send("No book found")`

server.js
```diff
@@ -1,30 +1,60 @@
...
+  const result = {numBooks: resultPage.length, 
+  page : page,
+  maxPage: maxPage,
+  books: resultPage}
+  res.json(result)
+})
+
+app.get('/books/id/:id', (req, res) => {
+  const id = req.params.id
+  const bookId = data.find((item)=>item.bookID == id)
```
HIPPIEKICK, line 39: You should use three plus signs when making checks like this, to make sure it's the same type as well. To make sure the id is a number (as we now the bookID is), you can set a plus sign in front of it to convert it to a number. If you want to read more, you can google "Unary plus".

`const bookId = data.find((item)=>item.bookID === +id)`

HIPPIEKICK: And the variable should be named after what it contains, which in this case is the whole book object and not just the id. So maybe just call it `book` since it's returning a book :)

server.js
```diff
@@ -1,30 +1,60 @@
...
+  res.json(bookId)
+  }
+  else {
+    res.send("404 not found")
+  }
+})
+
+app.get('/books/authors/:author', (req, res) => {
+  const author = req.params.author
+  const booksAuthor = data.filter((item)=>item.authors === author)
```
HIPPIEKICK, line 50: If you use `includes()` here instead of `filter()` you give the user the possibility to search for "J.K. Rowling" instead of exact match: "J.K. Rowling-Mary GrandPré"

server.js
```diff
@@ -1,30 +1,60 @@
...
+  res.json(result)
+})
+
+app.get('/books/id/:id', (req, res) => {
+  const id = req.params.id
+  const bookId = data.find((item)=>item.bookID == id)
+  if (bookId) {
+  res.json(bookId)
+  }
+  else {
```
HIPPIEKICK, line 43: No biggie but this looks better if it's on the same row:
`} else {`

PR #50, opened 2020-05-06
Project Express API - Mathias T

Nesiht:
This was fun, was looking forward to the backend!

### DIFF
--- a/server.js
+++ b/server.js
@@ -1,28 +1,14 @@
 import express from 'express'
 import bodyParser from 'body-parser'
 import cors from 'cors'
-
-// If you're using one of our datasets, uncomment the appropriate import below
-// to get started!
-// 
-// import goldenGlobesData from './data/golden-globes.json'
-// import avocadoSalesData from './data/avocado-sales.json'
 import booksData from './data/books.json'
-// import netflixData from './data/netflix-titles.json'
-// import topMusicData from './data/top-music.json'
 
-// Defines the port the app will run on. Defaults to 8080, but can be 
-// overridden when starting the server. For example:
-//
-//   PORT=9000 npm start
 const port = process.env.PORT || 8080
 const app = express()
 
-// Add middlewares to enable cors and json body parsing
 app.use(cors())
 app.use(bodyParser.json())
 
-// Start defining your routes here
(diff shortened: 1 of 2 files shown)


server.js
```diff
@@ -1,30 +1,75 @@
...
+
+  if(minPages) {
+    res.json(booksByMinPages)
+  } 
+  if(maxPages) {
+    res.json(booksByMaxPages)
+  }
+  
+  res.json(booksData)
+})
```
HIPPIEKICK, line 37: This is awesome and super useful for the user! It would be more RESTful to have this as a part of the `/books` endpoint but still keep both queries.

What happens if I want to search for a book with between 200 and 300 pages? An idea is to add a variable that you can refilter depending on the queries, and then in the end, just returning what's left:

```
let booksByPages = booksData

if(minPages) {
  booksByPages = booksByPages.filter((item) => item.num_pages >= minPages)
} 
if(maxPages) {
  booksByPages = booksByPages.filter((item) => item.num_pages <= maxPages)
}

res.json(booksByPages)

```

server.js
```diff
@@ -1,30 +1,75 @@
...
+app.get('/rating/:minRating', (req, res) => {
+  const minRating = req.params.minRating
+  let booksByRating = booksData.filter((item) => item.average_rating >= minRating)
+
+  if(booksByRating.length > 0) {
+    res.json(booksByRating)
+  } else {
+    res.status(404).json({ message: `Error, no books found with min rating of ${minRating}` })
+  }
 })
```
HIPPIEKICK, line 73: It would be nice if this also was a filter on the `/books` endpoint, e.g. `/books?minrating=5`

server.js
```diff
@@ -1,30 +1,75 @@
...
+
+// Filter books by ID
+app.get('/books/id/:id', (req, res) => {
+  const id = req.params.id
+  let booksById = booksData.filter((item) => item.bookID === +id)
+
+  if( booksById.length > 0 ) {
+    res.json(booksById)
+  } else {
+    res.status(404).json({ message: `Error, no book with id: ${id} found!` })
```
HIPPIEKICK, line 47: ⭐

PR #52, opened 2020-05-06
Malins project express

malin798:


filters.js
```diff
@@ -0,0 +1,36 @@
...
+
+  return movieList
+}
+
+export const filterMoviesByActor = (cast, movieList) => {
+  if (cast) {
+    return movieList.filter((item) => item.cast.includes(cast))
+  }
+  return movieList
+}
```
KarolCislowski, line 36: All three functions are doing the same, you can make it one function and just accept different arguments :)

server.js
```diff
@@ -1,33 +1,97 @@
 import express from 'express'
 import bodyParser from 'body-parser'
 import cors from 'cors'
+import netflixData from './data/netflix-titles.json'
+import { filterMoviesOnYear, filterMoviesByCountry, filterMoviesByActor, mappedNetflix } from './filters'
```
KarolCislowski, line 5: GREAT!! Utility functions extracted to separate file makes the server much easier to maintain :)

server.js
```diff
@@ -1,33 +1,97 @@
...
 
-// Add middlewares to enable cors and json body parsing
 app.use(cors())
 app.use(bodyParser.json())
 
-// Start defining your routes here
-app.get('/', (req, res) => {
-  res.send('Hello world')
+app.get('/', (request, response) => {
+  response.sendFile(path.join(__dirname+'/index.html'))
```
KarolCislowski, line 15: Great! Rendering the static file as the homepage is a great idea :)
If you want to use some dynamic data next time you can learn about the templating engines like ejs.
https://ejs.co/
It is an easy, nice and fun alternative to React if you have a small project no the board:)

server.js
```diff
@@ -1,33 +1,97 @@
...
-// Start the server
 app.listen(port, () => {
   console.log(`Server running on http://localhost:${port}`)
 })
+
+
+
+
+
+
```
KarolCislowski, line 97: Look to your vscode settings and maybe you will find trimming function which will cut the tail of the blank lines on the end of your file ;)

PR #55, opened 2020-05-06
Project 17 - Express API - Johanna Åstrand

jastrand:
This project was fun to explore the other side of an API. 

Things I would like to get clarified: 
I don't think my error when id is not found (row 39) is fully working, any ideas what I am doing wrong here? I would also like to know how to best handle if "type" is either TV Show or Movie. 

Thank you!!

### DIFF
--- a/server.js
+++ b/server.js
@@ -1,50 +1,65 @@
+
+
 import express from 'express'
 import bodyParser from 'body-parser'
 import cors from 'cors'
 
-// If you're using one of our datasets, uncomment the appropriate import below
-// to get started!
-// 
-import goldenGlobesData from './data/golden-globes.json'
+// // If you're using one of our datasets, uncomment the appropriate import below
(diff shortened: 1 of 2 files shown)


server.js
```diff
@@ -24,10 +14,54 @@ app.use(bodyParser.json())
...
+
+app.get('/titles/:id', (req, res) => {
+  const id = req.params.id
+  const titleId = netflixData.filter((item) => item.show_id === +id)
+  if (!titleId) {
+    res
+      .status(404)
+      .send({ error: `No movie or tv show found with id: "${id}" found` });
+  }
+  res.json(contentWithId);
```
KarolCislowski, line 44: I think you wanted to respond with `res.json(titleId)` or forget to create `contentWithId` ;)

jastrand: Done!

PR #57, opened 2020-05-06
Camilla Express API

camillagaus:


server.js
```diff
@@ -1,33 +1,45 @@
...
-  res.send('Hello world')
+  res.send('Netflix data. You can find all movies using /data, titles using /title, movies from a specific year with /year/[year] and a specific movie using /movie/[id]')
 })
 
-// Start the server
+app.get('/data', (req, res) => {
+  res.json(netflixData)
+})
+
+app.get('/title', (req, res) => {
```
KarolCislowski, line 20: RESTful endpoints should be in plural form:) so make it /titles /years /movies, please :)

PR #62, opened 2020-05-07
Project express api

Jenniferot:
I liked this project, nice to start from scratch and I look forward to next backend project. 

I'm not sure I understood exactly how to present the different endpoints and querys, but I got it down. One 404 message as well, just to see how it works, if the title of a show isn't right.

https://project-express-api-netflix.herokuapp.com/


server.js
```diff
@@ -24,10 +16,58 @@ app.use(bodyParser.json())
 
 // Start defining your routes here
 app.get('/', (req, res) => {
-  res.send('Hello world')
+  res.send('Endpoints = /shows & /titles/title & year/year & cast/cast & genre/listed_in & country/country&type=Movie or TV Show')
```
KarolCislowski, line 19: You can create an HTML file and render it here to make it look nicer, or use the template engine like ejs: https://ejs.co/

server.js
```diff
@@ -24,10 +16,58 @@ app.use(bodyParser.json())
...
+})
+
+app.get('/shows', (req, res) => {
+  res.json(data)
+})
+
+app.get('/titles/:title', (req, res) => {
```
KarolCislowski, line 26: the good REST api provides both: 

1. the list of something on path: `something/`
2. the single elements details on path: `something/:title`, `something/:id etc`

Here you have list of shows in `/shows` and the details under `/titles/:title`, you should make it `shows/:title`  or add separate endpoint `/titles` to list for example only titles from the data set.

PR #66, opened 2020-05-08
Week 17 Build an API

VickBe:
It went ok. I thought I got a hang of it quickly but then when I started making queries it got a bit trickier. 
If I would have had more time I would have liked to make the error message work and also there is a piece of code containing two queries. I don't get why they are not working. 
I would love for someone to have a look at the  pieces of code that I have marked with not woking and give feedback. I want understand why!!!! :)


server.js
```diff
@@ -27,6 +17,123 @@ app.get('/', (req, res) => {
   res.send('Hello world')
 })
 
+
+//GETS THE WHOLE DATA-SET UNFILTERED
+app.get('/collection', (req, res) => {
```
HIPPIEKICK, line 22: A more descriptive name would be `/shows` because that's what it's returning :)

server.js
```diff
@@ -27,6 +17,123 @@ app.get('/', (req, res) => {
...
+  let moviesFromYear = onlyMovies.filter((item) => item.release_year === +year)
+  const country = req.query.country
+
+  if (country) {
+    console.log(country)
+    moviesFromYear = moviesFromYear.filter((item) => item.country.toLowerCase() === country.toLowerCase())
+  }
+
+  res.json(moviesFromYear)
+})
```
HIPPIEKICK, line 48: Path params are mostly used to identify a specific resource, so I think it would be better to have this as a query param, since you're filtering here. Like this: `/movies?year=2019`

server.js
```diff
@@ -27,6 +17,123 @@ app.get('/', (req, res) => {
...
+//   const typeOfContent = req.query.type
+//   const showCountry = req.query.country
+
+//   if(typeOfContent) {
+//     collection = collection.filter((item) => item.type.toLowerCase() === typeOfContent.toLowerCase())
+//   }if(showCountry){
+//     collection = collection.filter((item) => item.country.toLowerCase() === showCountry.toLowerCase())
+//   }
+//   res.json(collection)
+// })
```
HIPPIEKICK, line 65: Did you have both this and this:
```
//GETS THE WHOLE DATA-SET UNFILTERED
app.get('/collection', (req, res) => {
  res.json(netflixData)
})
```
at the same time, I think that's the reason why. Just delete the other one and comment this back in. If the user doesn't write any queries it will return an unfiltered result 👍

server.js
```diff
@@ -27,6 +17,123 @@ app.get('/', (req, res) => {
...
+  const genre = req.params.genre
+  let onlyTvShows = netflixData.filter((item) => item.type === 'TV Show')
+  let specificGenreTv = onlyTvShows.filter((item) => item.listed_in.toLowerCase().includes(genre.toLowerCase()))
+  const year = req.query.year
+
+  if (year) {
+    specificGenreTv = specificGenreTv.filter((item) => item.release_year === +year)
+  }
+  res.json(specificGenreTv)
+})
```
HIPPIEKICK, line 97: These routes are named the same as well. If you want to filter on multiple params, it's better to use query params. Here's a stack overflow question with an example: [https://stackoverflow.com/c/technigo/questions/1101](https://stackoverflow.com/c/technigo/questions/1101)

server.js
```diff
@@ -27,6 +17,123 @@ app.get('/', (req, res) => {
...
+    title.includes({ specTitle }) === true
+
+    console.log(specTitle)
+    res.json(specTitle)
+  }
+  else {
+    res.status(404).json({ message: `${title} is not included in the Netlix collection` })
+
+  }
+})
```
HIPPIEKICK, line 117: This could also be a part of the `/shows` or `/collection` endpoint.

VickBe:
Thank you for really good comments! actually started from the beginning again and I got it to work in what I think is the "right" way. I've even got the error message to work. Had overcomplicated everything :) And I though returning a title was to return a single result but I understand now that you wanted us to use the id

HIPPIEKICK:
Looks super! Just be aware that when you're handling error like this, you're saying that if the user is not writing any queries - you're not going to return a result.
```
else {
    res.status(404).json({ error: 'not found' })
  }
```
If you write like this instead - you're still returning the data unfiltered:
```
if (shows.length === 0) {
  res.status(404).send(`Couldn't find any shows matching your search`)
} else {
  res.json(shows)
}
```

VickBe:
Ah! Thanks!! But can I just add that to my other if statements?

Den fre 15 maj 2020 kl 13:36 skrev Matilda Arvidsson <
notifications@github.com>:

HIPPIEKICK:
Yep! Either like the suggestion above, or the other way around:
```
if (shows.length > 0) //return shows
else //return error
```

PR #67, opened 2020-05-08
API express & Node

elinaarenstrup:


server.js
```diff
@@ -1,33 +1,68 @@
...
+app.use(cors());
+app.use(bodyParser.json());
+
+//Displays all awards
+app.get("/", (req, res) => {
+  res.json(data);
+});
+
+// Displays all the movies/shows
+app.get("/nominations", (req, res) => {
```
KarolCislowski, line 29: Why `/nominations`? Wouldn't it be better to use `/shows`?

server.js
```diff
@@ -1,33 +1,68 @@
...
+
+  res.json(showCountry);
+});
 
-// Start defining your routes here
-app.get('/', (req, res) => {
-  res.send('Hello world')
-})
+//Displays a movie/show with a specific show_id
+app.get("/showId/:showId", (req, res) => {
```
KarolCislowski, line 59: Because all your shows are listed in `/nominations`, so to be 100% REST this path to one show should be `/nominations/:showId`.

So if you like to change nominations to shows to be less golden globe and more Netflix ;) you should make this path `/shows/showId`

server.js
elinaarenstrup: I first went for Golden globes and then changed to Netflix. Seems I forgot to change the names lol 🙈

PR #68, opened 2020-05-08
project-express-api completed - Elsa Reuterswärd

elsareutersward:
View project:
https://er-project-express-api.herokuapp.com/

It was fun to start learning some backend. The aim of the project felt quite straightforward so I didn't have a lot of struggles. I sadly didn't have a lot of time to put on the project, so I feel that I might have done too little. I would have liked maybe do a simple frontend, used correct error codes or practiced some other methods for manipulating the data.


server.js
```diff
@@ -1,33 +1,117 @@
...
+  } else {
+    res.json(showPage)
+  }
+})
+
+app.get('/pages/:numpages/:page', (req, res) => {
+  const numpages = req.params.numpages
+  const page = +req.params.page - 1
+  let numberOfPages = booksData.filter((item) => item.num_pages <= +numpages)
+  let showPage = numberOfPages.slice(page*20, page*20+20)
```
KarolCislowski, line 105: You typed the same code 3 times. Wouldn't it be much nicer to create a separate function and call it wherever it is needed?

server.js
```diff
@@ -1,33 +1,117 @@
...
+  res.json(specificBook)
+
+  if (specificBook.length < 1) {
+    res.send('Could not find the requested book')
+  } else {
+    res.json(specificBook)
+  }
+})
+
+app.get('/ratings/:rating/:page', (req, res) => {
```
KarolCislowski, line 75: Nice structure, but with page numbers stored as URL query would be even nicer :)

PR #69, opened 2020-05-08
Olle Alm - Project Express Api

ollealm:
View project:
https://olle-api-books.herokuapp.com/

Fun to learn some back-end, and also practice some array functions.
Would like to add a simple front end if I get some extra time.

### DIFF
--- a/server.js
+++ b/server.js
@@ -19,7 +19,7 @@ app.use(bodyParser.json())
 
 // Root
 app.get('/', (req, res) => {
-  res.send('<code>Paths<br>"/books"<br>"/books/:id<br>"/isbn/:isbn"<code>')
+  res.send('<code>Paths:\n/books\n/books/:id\n/isbn/:isbn\n\nQueries:\n<code>')
 })
 
 
--- a/server.js
+++ b/server.js
@@ -15,11 +15,9 @@ app.use(cors())
 app.use(bodyParser.json())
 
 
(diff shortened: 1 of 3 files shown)


server.js
```diff
@@ -22,11 +14,98 @@ const app = express()
...
+        return 0;
+      });
+    } else if (sort === 'title') {
+      filteredBooks = filteredBooks.sort((a, b) => {
+        const titleA = a.title.toString().toLowerCase()
+        const titleB = b.title.toString().toLowerCase()
+        if (titleA < titleB) return -1;
+        if (titleA > titleB) return 1;
+        return 0;
+      });
```
KarolCislowski, line 69: Two times the same logic, so maybe it would nicer to have it in a separate function and pass author and title as props?

server.js
```diff
@@ -22,11 +14,98 @@ const app = express()
...
+    } else if (sort === 'pages') {
+      filteredBooks = filteredBooks.sort((a, b) => b.num_pages - a.num_pages);
+    } else if (sort === 'rating') {
+      filteredBooks = filteredBooks.sort((a, b) => b.average_rating - a.average_rating);
+    }
```
KarolCislowski, line 74: Same here:)

server.js
ollealm: Good point, will do! Thanks!

PR #70, opened 2020-05-08
Express-api - Josephine

prosperpine:
Learning this part was more fun and interesting than I thought it would be. :)
I did a simply front-end part as well, to see how it was to use an API that I made on my own and I think I learned a lot in the process. 

Link to deployed project: 
https://first-express-project.herokuapp.com/

Link to frontend repo: 
https://github.com/prosperpine/express-app
Deployed frontend: 
https://competent-beaver-d3f54d.netlify.app/

### DIFF
--- a/server.js
+++ b/server.js
@@ -33,6 +33,13 @@ app.get('/movies', (req, res) => {
   res.json(netflixData);
 });
 
+// route for getting data on a certain movie
+app.get('/movies/:movie', (req, res) => {
+  const movie = req.params.movie;
+  const chosenMovie = netflixData.filter((item) => item.title === movie);
+  res.json(chosenMovie);
+});
+
 // route for getting movies from a certain category
 app.get('/genres/:genre', (req, res) => {
   const genre = req.params.genre;
@@ -47,29 +54,12 @@ app.get('/year/:year', (req, res) => {
   res.json(releaseYear);
 });
 
-// Filter based on title AND/OR country
-// app.get('/shows', (req, res) => {
-//   // Query parameter
-//   const titleSearchString = req.query.title
-//   const countrySearchString = req.query.country
-
-//   let filteredShows = netflixData;
-
-//   if (titleSearchString) {
-//     filteredShows = filteredShows.filter(item => {
-//       const itemTitle = item.title.toString()
-//       return itemTitle.includes(titleSearchString)
-//     })
-//   }
-
-//   if (countrySearchString) {
-//     filteredShows = filteredShows.filter(item => {
-//       const itemCountry = item.country.toString()
-//       return itemCountry.includes(countrySearchString)
-//     })
-//   }
-
-//   res.json(filteredShows);
+// route for getting movies from a certain country
+app.get('/countries/:country', (req, res) => {
+  const country = req.params.country;
+  const chosenCountry = netflixData.filter((item) => item.country === country);
+  res.json(chosenCountry);
+});
 
 app.use(function (err, req, res, next) {
   console.error(err.stack);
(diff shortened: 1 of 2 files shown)


server.js
```diff
@@ -1,33 +1,85 @@
...
-app.use(bodyParser.json())
+app.use(cors());
+app.use(bodyParser.json());
 
 // Start defining your routes here
 app.get('/', (req, res) => {
-  res.send('Hello world')
-})
+  res.send(
+    'paths: movies, movies/movie, genres/genre, actors/actor, directors/director, year/year, countries/country'
```
HIPPIEKICK, line 22: Nice! ⭐ If you want this documentation formatted, you can take a look here: https://www.npmjs.com/package/express-list-endpoints

server.js
```diff
@@ -1,33 +1,85 @@
...
+});
+
+// route for getting data on a certain movie
+app.get('/movies/:movie', (req, res) => {
+  const movie = req.params.movie;
+  const chosenMovie = netflixData.filter(
+    (item) => item.title.toString().toLowerCase() === movie.toLowerCase()
+  );
+  res.json(chosenMovie);
+});
```
HIPPIEKICK, line 37: Hmm, I think you've done kind of a mix here. Using `filter()` you get an array with the result back. If this route is going to return a single result, you could rename it `/movies/:id` and then use `find()` to find an id and to make sure you only get a single result.

You could keep the filter and move it to the `/movies` endpoint but use it as a query param. Path params are mostly used to identify a specific resource, so when you just want to filter, you use query params.

Another thing to think about is that when you're using === the param has to be an exact match with the title in your data, so if I just write "Irishman" it won't be a match to "The Irishman". To give the user the possibility to search for titles including a specific word you can use `includes()`.

```
const movie = req.query.movie;

const chosenMovie = netflixData.filter((item) =>
    item.title.toString().toLowerCase().includes(movie)
```

Then it would find "The Irishman", even if the user just searches for "Irishman".

server.js
```diff
@@ -1,33 +1,85 @@
...
+
+// route for getting movies from a certain category
+app.get('/genres/:genre', (req, res) => {
+  const genre = req.params.genre;
+  const movieGenre = netflixData.filter((item) =>
+    item.listed_in.toLowerCase().includes(genre.toLowerCase())
+  );
+  res.json(movieGenre);
+});
```
HIPPIEKICK, line 46: It would be nice if this also was a filter on the `/movies` endpoint, e.g. `/movies?genre=drama`

server.js
```diff
@@ -1,33 +1,85 @@
...
+
+// route for getting movies from a certain country
+app.get('/countries/:country', (req, res) => {
+  const country = req.params.country;
+  const chosenCountry = netflixData.filter(
+    (item) => item.country.toLowerCase() === country.toLowerCase()
+  );
+  res.json(chosenCountry);
+});
```
HIPPIEKICK, line 80: You could use many query parameters instead of creating new routes. Routes should be named after the data they are returning. Even though you sort the data on actor, director, year or country - they still give you back movies. So all of these routes could be params for the `/movies` endpoint. Take a look at this stack overflow question to see how you can add multiple queries: [https://stackoverflow.com/c/technigo/questions/1101](https://stackoverflow.com/c/technigo/questions/1101)

HIPPIEKICK:
> Learning this part was more fun and interesting than I thought it would be. :)
> I did a simply front-end part as well, to see how it was to use an API that I made on my own and I think I learned a lot in the process.
> 

I really agree with you on that backend is fun! And so nice to see that you did a frontend as well ⭐

prosperpine:
I have added an endpoint for specific id now.

prosperpine:
Hi!
Yes, I've made the change now!
BR,

Josephine

tors 14 maj 2020 kl. 16:05 skrev Matilda Arvidsson <notifications@github.com

HIPPIEKICK:
Awesome! Approving ⭐

PR #83, opened 2020-12-07
Book API - Annika Melin

ankimelin:
This was fun!! Especially to see many of the JavaScript methods coming to life. I even had to push myself using Regex at one point - and it made sense to do so! So much easier to understand its right to exist after having used it in a project.

Live API: https://project-book-api.herokuapp.com/
Endpoint documentation in readme file.

### DIFF
--- a/server.js
+++ b/server.js
@@ -2,32 +2,118 @@ import express from 'express'
 import bodyParser from 'body-parser'
 import cors from 'cors'
 
-// If you're using one of our datasets, uncomment the appropriate import below
-// to get started!
-// 
-// import goldenGlobesData from './data/golden-globes.json'
-// import avocadoSalesData from './data/avocado-sales.json'
-// import booksData from './data/books.json'
-// import netflixData from './data/netflix-titles.json'
-// import topMusicData from './data/top-music.json'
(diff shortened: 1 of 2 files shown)


server.js
```diff
@@ -2,32 +2,162 @@ import express from 'express'
...
+        return a.num_pages - b.num_pages
+      })
+      const shortestBooks = sortedBooksOnPages.slice(0, shortest)
+      return response.json(shortestBooks)
+    }
+  }
+
+  else {
+    return response.json(booksData.slice(0, 20))
+  }
```
HIPPIEKICK, line 117: Instead of having all these if/else, what do you think about refiltering the same data?

```
let filteredBooks
if (author) {
  filteredBooks  = //filtered on author
}
if (language) {
  filteredBooks = //filtered on language
}
```
and so on? :)

HIPPIEKICK (approved):
Great job Annika! A lot of functionality and nice to see that you tried out regular expressions as well. The documentation is awesome, [here](https://www.npmjs.com/package/express-list-endpoints)'s a cool lib that you can try out next time if you'd like to include the docs in your API. Keep up the good work!

PR #91, opened 2020-12-11
Project-express-api - Ellen

ellenbjo:
It has been a fun week! It was nice to focus more on vanilla js. I think I needed that practice! 

https://ellens-express-deployment.herokuapp.com/

### DIFF
--- a/server.js
+++ b/server.js
@@ -1,6 +1,7 @@
 import express from 'express'
 import bodyParser from 'body-parser'
 import cors from 'cors'
+import listEndpoints from 'express-list-endpoints'
 
 import booksData from './data/books.json'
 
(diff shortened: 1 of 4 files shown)


server.js
```diff
@@ -23,10 +13,79 @@ app.use(cors())
 app.use(bodyParser.json())
 
 // Start defining your routes here
+// request is incoming an res is outgoing
+// used listEndpoints to list all routes
+
+const ERROR_MESSAGE = 'No such book found'
```
EmOlsz, line 19: Good idea with predefined error message! Also, I'm happy to see proper naming convention (uppercase letters divided by `_`)

EmOlsz (approved):

PR #90, opened 2020-12-11
project-express-api-ThereseLinden

thereselinden:
I'm happy with this weeks project, even though I had not much time to actually do some coding. 
I'm glad that I managed to create a POST request, it worked in postman so I hope I created that endpoint correctly. 

If I had more time I would have created some frontend that would hold the documentation part.
Right now I'm using express list endpoints package, and I don't think that is very descriptive. 

I would like some information on other documentation tools that can be used, how to use them and so on.  E.g. swagger.

View it live: 
[https://therese-book-api.herokuapp.com/]

PR #98, opened 2020-12-11
Project-express-api Emelie Svensson

emeliesv:
This weeks project was a lot of fun! I feel like I still have a lot to learn on how to check if an API is RESTful or not, but I found that this checklist was helpful: https://tomyjaya.github.io/2018/02/01/api-design/

I aimed for some of the stretchgoals this week, but not all. Instead I decided to learn how to use the POST-request and update my API with some more input. I'm very pleased with the way that turned out. Further on I'd like to add a DELETE-request as well and also make some function that renders unique id-s for each input since there now is the possibility for multiple, identical id:s (that is an issue). 

The deployed API can be found at: https://project-api-books-emelie.herokuapp.com
(It doesn't have any documentation yet)


server.js
```diff
@@ -1,33 +1,120 @@
...
+  const otherlanguages = req.query.otherlang
+  const sorted = req.query.sort
+  
+  if (language) {
+    books = filterOnLanguage(books, language, ERROR_MSG, res)
+  } else if (otherlanguages){
+    books = filterOtherLanguages(books, language, ERROR_MSG, res)
+  }
+
+  if(sorted) {
```
HIPPIEKICK, line 93: Minor but, spacing: `if (sorted) {`

server.js
```diff
@@ -1,33 +1,120 @@
...
+    books = sortByNumPages(books, sorted)
+  }
+  
+  const currentPage = pagination(books, page)
+  res.json(currentPage)
```
HIPPIEKICK, line 98: Maybe this could be called `filteredBooks`

server.js
```diff
@@ -1,33 +1,120 @@
...
+})
+
+app.post('/books', (req, res) => {
+  const input = req.body
```
HIPPIEKICK, line 102: If you specified the input you will get, you could decide for yourself which fields should come from the user and which ones should be generated by you. [Here](https://github.com/Technigo/project-express-api/pull/90/files)'s an example from one of your classmates :)

server.js
```diff
@@ -1,33 +1,120 @@
...
+})
+
+app.get('/books/:book', (req, res) => {
+  const book = req.params.book
+  const result = booksData.find((item) => item.bookID === +book)
+  if (!result) {
+    res.status(404).json(ERROR_MSG)
+  }
+  res.json(result)
 })
```
HIPPIEKICK, line 116: Great with some error handling! A suggestion for the name of the path param would be `bookID` or just `id`

HIPPIEKICK (approved):
Really nice to see that you've managed to create pagination,  a `POST` request and refiltering of the data! Just give your variables name an extra round of thoughts and it will be even more impressive!

PS.

😂 👇 
Skärmavbild 2021-01-05 kl  17 58 23

PR #99, opened 2020-12-11
handing in project 17

HenrikeW:
When it came to code and syntax, this project was pretty easy. But it was interesting to think about how to make it most logic and truly RESTful. It's a hard topic I think and I'll need some time to try and experience different approaches.

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,6 +15,7 @@
     "body-parser": "^1.19.0",
     "cors": "^2.8.5",
     "express": "^4.17.1",
+    "express-list-endpoints": "^5.0.0",
     "nodemon": "^2.0.2"
   },
   "devDependencies": {
--- a/server.js
+++ b/server.js
@@ -1,6 +1,7 @@
-import express from 'express'
-import bodyParser from 'body-parser'
-import cors from 'cors'
+import express from 'express';
+import bodyParser from 'body-parser';
+import cors from 'cors';
+import books from './data/books.json';
 
 // If you're using one of our datasets, uncomment the appropriate import below
 // to get started!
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -15,6 +15,7 @@
     "body-parser": "^1.19.0",
     "cors": "^2.8.5",
     "express": "^4.17.1",
+    "express-list-endpoints": "^5.0.0",
```
HIPPIEKICK, line 18: Yey! 💯

server.js
```diff
@@ -15,19 +17,62 @@ import cors from 'cors'
...
-app.get('/', (req, res) => {
-  res.send('Hello world')
-})
+const endPoints = require('express-list-endpoints')
+app.get("/", (request, response) => {
+  response.send(endPoints(app));
+});
+
+// Get all the books in the list & filter options
+app.get('/books', (req, res) => {
```
HIPPIEKICK, line 34: Try to be consistent - choose req, res or request, response and go with the same for all routes

server.js
```diff
@@ -15,19 +17,62 @@ import cors from 'cors'
...
+  if (minRating) {
+    filteredBooks = filteredBooks.filter((book) => book.average_rating >= minRating)
+  };
+
+  if (author) {
+    filteredBooks = filteredBooks.filter((book) => book.authors.toLowerCase().includes(author.toLowerCase()))
+  };
+
+  res.json(filteredBooks);
+});
```
HIPPIEKICK, line 59: Really nice with refiltering!

server.js
```diff
@@ -15,19 +17,62 @@ import cors from 'cors'
...
+// Get all the books in the list & filter options
+app.get('/books', (req, res) => {
+  // get only books that have been ranked
+  const isRanked = req.query.isRanked
+
+  // get only books with a certain minimum ranking
+  const minRating = req.query.minRating;
+
+  // get only books by a certain author
+  const author = req.query.author;
```
HIPPIEKICK, line 42: This could be destructured, like:
`const { isRanked, minRating, author } = req.query;`

server.js
```diff
@@ -15,19 +17,62 @@ import cors from 'cors'
...
+  const book = books.find((book) => book.bookID === +id);
+  res.json(book);
+});
+
+// Get all books in a specific language
+app.get('/languages/:language', (req, res) => {
+  const language = req.params.language;
+  const booksInLanguage = books.filter((book) => book.language_code === language);
+  res.json(booksInLanguage);
+});
```
HIPPIEKICK, line 73: Would be nice if this too were a query under the `/books` endpoint, since it returns books and not languages ;)

HIPPIEKICK (approved):
Good job Henrike! Great that you found the listing endpoints lib and that you tried out some querying 💯

PR #102, opened 2020-12-11
project-express-api - Jamie Cook

the-j-curl:
Link to my deployed API: https://jamie-books-api.herokuapp.com/

Reflections in this week: I thought this was a fun project to start to learn about backend, but it felt quite basic. I enjoyed learning about how to create endpoints and query's/parameters. I think I will enjoy it a lot more when we get to connect it to the frontend.

Things I want clarified: I would like to practice working with other datasets, and also know what other things are possible with the backend, apart from just queries and params.


server.js
```diff
@@ -1,33 +1,98 @@
...
+// Add middlewares to enable cors and json body parsing - makes it easier for frontend to use the API. Corrs allows API's to say where requests can come from. To add extra security
+app.use(cors());
+app.use(bodyParser.json()); //This allows Express to read JSON in POST requests
 
 // Start defining your routes here
-app.get('/', (req, res) => {
-  res.send('Hello world')
-})
+app.get("/", (req, res) => {
+  res.send(listEndpoints(app));
```
EmOlsz, line 23: Great job :)

server.js
```diff
@@ -1,33 +1,98 @@
...
+});
+
+app.get("/books", (req, res) => {
+  let page = req.query.page ?? 1;
```
EmOlsz, line 27: I'm impressed by seeing `??` - sometimes called nullish operator. Just to be sure you are aware of all features this operator offers and difference between `??` and `||`, check out https://www.javascripttutorial.net/es-next/javascript-nullish-coalescing-operator/

server.js
```diff
@@ -1,33 +1,98 @@
...
+  const pageSize = req.query.pageSize ?? 20; // this allows you to query page numbers, example: http://localhost:8080/books?page=1&pageSize=10
+  page = +page - 1;
+  let bookList = booksData;
+  console.log(`page = ${page}`);
+  const startIndex = page * pageSize; //calculate the start index
+  const endIndex = startIndex + +pageSize;
+  const booksForPage = bookList.slice(startIndex, endIndex); //Slice creates a new array to limit the number of books returned to 20 per page
+
+  if (bookList.length === 0) {
```
EmOlsz, line 36: or just `if (!bookList.length)`

server.js
```diff
@@ -1,33 +1,98 @@
...
+    res.status(404).send(ERROR_NOT_FOUND);
+  }
+  const returnObject = {
```
EmOlsz, line 39: Very clean and solid data structure, well done!

server.js
```diff
@@ -1,33 +1,98 @@
...
+    pageSize: booksForPage.length,
+    currentPage: page + 1,
+    maxPages: Math.ceil(bookList.length / pageSize),
+    totalNumOfBooks: bookList.length,
+    results: booksForPage,
+  };
+  res.send(returnObject);
+});
+
+app.get("/books/bookauthor", (req, res) => {
```
EmOlsz, line 49: It would be more restfull to create this endpoint with normal parameter :

```
app.get("/books/bookauthor/:author", (req res) => { ... });
```

Think about query parameteres as about something optional - we can provide it in URL, but we don't have to.
Since this endpoint is created for book authors only, it's reasonable to build it with normal parameter, like suggested above

server.js
```diff
@@ -1,33 +1,98 @@
...
+    );
+    if (bookListAuthor.length === 0) {
+      res.status(404).send(ERROR_NOT_FOUND);
+    }
+    res.json(bookListAuthor);
+  }
+});
+
+app.get("/books/booktitle", (req, res) => {
+  const { title } = req.query;
```
EmOlsz, line 65: Same goes for this one

EmOlsz (approved):

PR #107, opened 2020-12-12
Express-API - Anna Hellqvist

AnnaHellqvist:
This week it has been fun to get an insight in the backend. It has also been quite challenging. 
I felt that it was hard to grasp the RESTful part of this week. What Restful is and isn't, I would have appreciated a few examples to compare to make it more clear.

### DIFF
--- a/server.js
+++ b/server.js
@@ -54,7 +54,18 @@ app.get('/rating/:rating', (request, response) => {
 app.get('/author/:author', (request, response) => {
   const author = request.params.author;
   const booksByAuthor = booksData.filter((book) => book.authors === author)
+
   response.json(booksByAuthor)
+})
+
+// returns the ten books with highest rating
+app.get('/top/:number', (request, response) => {
+  const number = request.params.number;
+  const sortedBooks = [...booksData];
+  sortedBooks.sort((a, b) => b.average_rating - a.average_rating)
+  const topTenBooks = sortedBooks.slice(0, number);
+
+  response.json(topTenBooks)
 })
 
 // filtrera information ? blir mer än en i resultat
--- a/server.js
+++ b/server.js
@@ -3,15 +3,6 @@ import bodyParser from 'body-parser'
 import cors from 'cors'
 import booksData from './data/books.json'
 
-// If you're using one of our datasets, uncomment the appropriate import below
-// to get started!
-// 
-// import goldenGlobesData from './data/golden-globes.json'
-// import avocadoSalesData from './data/avocado-sales.json'
-
-// import netflixData from './data/netflix-titles.json'
-// import topMusicData from './data/top-music.json'
-
 // Defines the port the app will run on. Defaults to 8080, but can be 
 // overridden when starting the server. For example:
 //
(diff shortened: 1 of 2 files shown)


server.js
```diff
@@ -1,33 +1,73 @@
-import express from 'express'
+import express, { response } from 'express'
```
HIPPIEKICK, line 1: `response` should be a part of `express`, did you have trouble with using it without specifying it?

server.js
```diff
@@ -1,33 +1,73 @@
...
 app.use(cors())
 app.use(bodyParser.json())
 
 // Start defining your routes here
-app.get('/', (req, res) => {
-  res.send('Hello world')
+// accepting get requests to this endpoint, / means no path after the address
+app.get('/', (request, response) => {
+  response.send('Welcome to my books API // Anna Hellqvist 📚')
+})
```
HIPPIEKICK, line 21: [Here](https://www.npmjs.com/package/express-list-endpoints)'s a cool lib that you can try out next time if you'd like to include some documentation in your API!

server.js
```diff
@@ -1,33 +1,73 @@
...
+
+// retrieves the whole books array
+app.get('/books', (request, response) => {
+
+  const { author, toplist, rating, language } = request.query;
```
HIPPIEKICK, line 26: camelCase ;)

server.js
```diff
@@ -1,33 +1,73 @@
...
+    } else {
+      response.json({ error: 'No books with that rating were found' })
+    }
+  } else if (language) {
+    const filteredLanguage = booksData.filter((book) => book.language_code === language);
+    response.json(filteredLanguage);
+  } else {
+    response.json(booksData);
+  }
+})
```
HIPPIEKICK, line 52: Instead of having all these if/else, what do you think about refiltering the same data? Wouldn't it be nice if the user was able to query multiple things? 😇 Something like:

```
let filteredBooks
if (author) {
  filteredBooks  = //filtered on author
}
if (language) {
  filteredBooks = //filtered on language
}
res.json(filteredBooks)
```
and so on

HIPPIEKICK: Also case insensitivity and partial query would be nice

server.js
```diff
@@ -1,33 +1,73 @@
...
+
+//retrieves one book based on ID
+app.get('/books/:id', (request, response) => {
+  const { id } = request.params;
+  const bookFound = booksData.find(book => book.bookID === +id);
+
+  bookFound ? response.json(bookFound) : response.json({ error: 'No book was found' })
+})
```
HIPPIEKICK, line 60: Great with some error handling 💯

HIPPIEKICK: And that you're consistent with the destructuring 👍

server.js
```diff
@@ -1,33 +1,73 @@
...
+
+//------------------------------------------------------------------------------------------------------
+
+// empty/dummy endpoints
+app.get('/books/:year', (request, response) => {
+  // This should return all the books released in a specific year
 })
```
HIPPIEKICK, line 67: This too could be a query under the `/books` endpoint:
`/books?year=2010`

HIPPIEKICK (approved):
Good job Anna! You seem to have figured it out well - endpoints should be named after what they return and not after how the result is filtered or sorted etc. Meaning - `/books` and `/books/:id` are top notch naming 👍

server.js
AnnaHellqvist: I can't remember adding this one, I think vscode adds things by itself sometimes 😝

server.js
AnnaHellqvist: Thank you, I will check it out! ✨

PR #110, opened 2020-12-13
Peggy's Express API

pwangy:
Link to project: 
https://pw-c19-api.herokuapp.com/

Reflections on this week's project: 
This week I challenged myself to work from a dataset of my own choosing. After pulling the data and narrowing down the Covid-19 data to a selection of countries/provinces I made sure that every row of data had it's own unique ID. Then I had to determine what endpoints would make the most sense to someone who might be interested in using this data. After creating a handfull of endpoints and reaching all the Blue Level requirements, I spent a little time thinking about what the frontend might look like but I decided against merge this work into the master branch since it was left unfinished. 
Instead, the focus of my week was on catching up with the Career-Coaching sprints and networking before the holiday break. 

No requests for further clarification this week.


server.js
```diff
@@ -2,32 +2,59 @@ import express from 'express'
...
-  console.log(`Server running on http://localhost:${port}`)
+// Route for entire array
+app.get('/all', (req, res) => {
+  res.json(c19)
+})
+
+// Route to get data for a specific Type and location
+// Returns a single item from the array
+app.get('/all/:id', (req,res) => {
+  const id = req.params.id
```
EmOlsz, line 31: Or just `const { id } = req.params;`

server.js
```diff
@@ -2,32 +2,59 @@ import express from 'express'
...
+  let byID = c19.filter((item) => item.id === id)
```
EmOlsz, line 32: No need to use `let`, it should be `const`

server.js
```diff
@@ -2,32 +2,59 @@ import express from 'express'
...
+  res.json(byID)
+})
+
+// Route to get data by country
+// Returns all 3 types of data for each country and any province defined for country
+app.get('/countries/:country', (req, res) => {
+  const country = req.params.country
+  let byCountry = c19.filter((item) => item.country === country)
```
EmOlsz, line 40: Same here

server.js
```diff
@@ -2,32 +2,59 @@ import express from 'express'
...
+  res.json(byCountry)
+})
+
+// Route to get data by Type: total Cases, Deaths, or Recovered
+// Route can also be filtered by either Region or Country
+app.get('/types/:type', (req,res) => {
+  const type = req.params.type
+  let byType = c19.filter((item) => item.type === type)
```
EmOlsz, line 48: And over here ;)

EmOlsz (approved):

PR #112, opened 2020-12-13
Golden Globe API using Express 

destinyleia:
See readme for more info!

### DIFF
--- a/server.js
+++ b/server.js
@@ -1,4 +1,4 @@
-import express, { response } from 'express'
+import express from 'express'
 import bodyParser from 'body-parser'
 import cors from 'cors'
 
@@ -25,23 +25,23 @@ app.get('/nominations', (request, response) => {
   response.json(goldenGlobesData)
 })
 
-app.get('/nominations/years/:year', (request, response) => {
+app.get('/year/:year', (request, response) => {
   const year = request.params.year
   const showWon = request.query.won
-  let nominationsFromYear = goldenGlobesData.filter((item) => item.year_award === +year)
+  let yearFromYear = goldenGlobesData.filter((item) => item.year_award === +year)
 
   if (showWon) {
-    nominationsFromYear = nominationsFromYear.filter((item) => item.win)
+    yearFromYear = yearFromYear.filter((item) => item.win)
   }
-  response.json(nominationsFromYear)
+  response.json(yearFromYear)
 })
 
 // Tried to get a uniqe category back by using this code but instead a got back an empty array... 
-app.get('/nominations/categorys/:category', (request, response) => {
+app.get('/nominations/:nominee', (request, response) => {
   const category = request.params.category
-  const nominationsFilm = goldenGlobesData.filter((item) => item.category === category)
+  const nominationsFilm = goldenGlobesData.filter.length((item) => item.category === category)
 
-  if (!nominationsFilm) {
+  if (nominationsFilm) {
     response.status(404).json("ERROR - This category doesn't exist in the golden-globes API")
   }
     response.json(nominationsFilm)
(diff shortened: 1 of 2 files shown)


server.js
```diff
@@ -1,32 +1,52 @@
...
+app.get('/nominations/years/:year', (request, response) => {
+  const year = request.params.year
+  const showWon = request.query.won
+  let nominationsFromYear = goldenGlobesData.filter((item) => item.year_award === +year)
+
+  if (showWon) {
+    nominationsFromYear = nominationsFromYear.filter((item) => item.win)
+  }
+  response.json(nominationsFromYear)
+})
```
HIPPIEKICK, line 37: Endpoints should be named after what they return and not after how they are filtered or sorted. This could therefore be a query param under the nominations endpoint:
`nominations?year=2010&win=true`

server.js
```diff
@@ -1,32 +1,52 @@
...
+
+// Tried to get a uniqe category back by using this code but instead a got back an empty array... 
+app.get('/nominations/categorys/:category', (request, response) => {
+  const category = request.params.category
+  const nominationsFilm = goldenGlobesData.filter((item) => item.category === category)
+
+  if (!nominationsFilm) {
```
HIPPIEKICK, line 44: An empty array is not a falsy value, you need to check the length if you want to make this distinction

server.js
```diff
@@ -1,32 +1,52 @@
...
+  let nominationsFromYear = goldenGlobesData.filter((item) => item.year_award === +year)
+
+  if (showWon) {
+    nominationsFromYear = nominationsFromYear.filter((item) => item.win)
+  }
+  response.json(nominationsFromYear)
+})
+
+// Tried to get a uniqe category back by using this code but instead a got back an empty array... 
+app.get('/nominations/categorys/:category', (request, response) => {
```
HIPPIEKICK, line 40: For an endpoint that returns a single result it's a good idea to use the ID. Since this data doesn't have any ID's included - try to think what other element in the data comes the closest to being unique (e.g. nominee?) and change this endpoint to find that (filter returns an array as you mentioned) or add some ID's to your data

HIPPIEKICK (changes requested):
From the requirements:
- A minimum of one endpoint to return a single result (single element).
Fix that and you're good to go ✨

server.js
destinyleia: Can you please check if I thought right with the solution of the problem? 👍🏽  
on line 42

server.js
destinyleia: A took away nominations and took year/:year instead :)

destinyleia:
Hi Matilda, I've updated my code now and hope it thought right now  :)

server.js
HIPPIEKICK: It's returning nominations, so a more suiting name would be `/nominations`. My suggestion was to have the year as a query param under your main endpoint (instead of having it as another endpoint :) )

server.js
HIPPIEKICK: You can have a look at the first answer to this [Stack overflow question](https://stackoverflow.com/questions/24403732/how-to-check-if-array-is-empty-or-does-not-exist) :)

server.js
```diff
@@ -2,31 +2,51 @@ import express from 'express'
...
+
+  if (showWon) {
+    yearFromYear = yearFromYear.filter((item) => item.win)
+  }
+  response.json(yearFromYear)
+})
+
+// Tried to get a uniqe category back by using this code but instead a got back an empty array... 
+app.get('/nominations/:nominee', (request, response) => {
+  const category = request.params.category
```
HIPPIEKICK, line 41: Also, the param doesn't correspond here!

HIPPIEKICK: This still needs to be fixed:

> From the requirements:
> 
> - A minimum of one endpoint to return a single result (single element).
> Fix that and you're good to go ✨

PR #113, opened 2020-12-13
Agnes spotify-api

aagneshansson:
Link to Heroku: https://dashboard.heroku.com/apps/agnes-tracks/
Link to front-end: https://github.com/aagneshansson/Agnes-frontend-api

### DIFF
--- a/server.js
+++ b/server.js
@@ -1,4 +1,4 @@
-import express from 'express'
+import express, { response } from 'express'
 import bodyParser from 'body-parser'
 import cors from 'cors'
 
@@ -24,17 +24,13 @@ app.use(cors())
 app.use(bodyParser.json())
 
 // Start defining your routes here
-app.get('/', (request /*Incoming*/, response /*Outgoing*/) => {
-  //Any code we want
-  //Database connections
-  //Data lookups
-  //Third third api request
-  response.send('Good morning, have a wonderful day')
+app.get('/', (request, response) => {
+  response.send('Welcome, /topmusic will show you the top tracks!')
 })
 
-// All top music 
-app.get('/topmusic', (req, res) => {
-  res.json(topMusicData)
+// The first endpoint returns a collection of all top music 
+app.get('/topmusic', (request, response) => {
+  response.json(topMusicData)
 })
 
 // All artists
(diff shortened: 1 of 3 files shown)


server.js
```diff
@@ -9,25 +9,112 @@ import cors from 'cors'
...
+
+  response.json(uniqueTracks);
+});
+
+// Search by id 
+app.get('/topmusic/:id', (request, response) => {
+  const id = request.params.id
+  const topmusicID = topMusicData.find((item) => item.id === +id);
+  response.json(topmusicID);
+});
```
HIPPIEKICK, line 57: Nice!

server.js
```diff
@@ -9,25 +9,112 @@ import cors from 'cors'
...
+
+  // If track is not found
+  if (uniqueTracks.length === 0) {
+    const error = new Error(`${trackName} not found`);
+    error.status = 404;
+    throw error;
+  };
+
+  response.json(uniqueTracks);
+});
```
HIPPIEKICK, line 74: This doesn't have to be its own endpoint, wouldn't it be nice if this was a query param under your first endpoint? Something like:
`/songs?trackName="china"`

Super nice with error handling! You could improve the search function by making it case insensitive as well as letting the user search for a part of a track's name :)

HIPPIEKICK: Also - wouldn't it be nice if the whole song object were returned ;)

server.js
```diff
@@ -9,25 +9,112 @@ import cors from 'cors'
...
+
+// Filter by popularity, /number will tell how many top songs to show
+app.get('/topmusic/:number', (request, response) => {
+  const number = request.params.number
```
HIPPIEKICK, line 78: Not sure what this number means. Could this be better named? 😇

server.js
```diff
@@ -9,25 +9,112 @@ import cors from 'cors'
...
+
+// Shows the top 3 most popular songs
+app.get('/top3', (request, response) => {
+  const three = request.params.three
+  const sortedSongs = [...topMusicData]
+  sortedSongs.sort((a, b) => b.popularity - a.popularity)
+  const topThreeSongs = sortedSongs.slice(0, 3)
+
+  response.json(topThreeSongs);
+});
```
HIPPIEKICK, line 94: This endpoint seems to do the same thing as the last one 😇  Also - `three` is not used

server.js
```diff
@@ -9,25 +9,112 @@ import cors from 'cors'
...
+    const topTracks = topMusicData.sort((a, b) => b[sort] - a[sort]);
+    response.json(topTracks)
+
+    if (topTracks.length === 0) {
+      const error = new Error("Not found");
+      error.status = 404;
+      throw error
+    };
+  };
+});
```
HIPPIEKICK, line 115: Wow, really nice with some sorting! Just remember that endpoints should be named after what they return and not on how the data is filtered or sorted, so this too could be a query under your first endpoint. Something like:
`/songs?sort=popularity`

HIPPIEKICK (changes requested):
Nice job! Make sure you're naming your endpoints RESTfully - I made some comments about it :) Change that and you're good to go!

server.js
```diff
@@ -1,33 +1,116 @@
...
+// All tracks
+app.get('/songs', (request, response) => {
+  const tracks = topMusicData.map(item => item.trackName);
+  const uniqueTracks = [...new Set(tracks)];
+
+  response.json(uniqueTracks);
+});
+
+// Sorting all tracks by popularity /toplist?sort=popularity
+app.get('/songs?sort=popularity', (request, response) => {
```
HIPPIEKICK, line 50: The query param should not be hardcoded in the endpoint :)

HIPPIEKICK (approved):

PR #115, opened 2020-12-13
Express Nathalie Nordström

nathalienordstrom:


server.js
```diff
@@ -1,32 +1,30 @@
...
-// Add middlewares to enable cors and json body parsing
 app.use(cors())
 app.use(bodyParser.json())
 
-// Start defining your routes here
 app.get('/', (req, res) => {
   res.send('Hello world')
 })
 
+app.get('/year', (req, res) => {
```
HIPPIEKICK, line 18: Endpoints should be named after what they return so this could be named `shows` or `/titles` or similar

server.js
```diff
@@ -1,32 +1,30 @@
...
+  res.json(data)
+})
+app.get('/years/:year', (req, res) => {
+  const year = req.params.year
+  const releaseYear = data.filter((item) => item.release_year === +year)
+
+    res.json(releaseYear)
+})
```
HIPPIEKICK, line 26: Nice with some filtering 👍 
A suggestion would be to make this a query param under your first endpoint, like:
`/shows?year=2020`

HIPPIEKICK (changes requested):
From the requirements:
- A minimum of one endpoint to return a single result (single element).

Create an endpoint that returns a single element (e.g. using ID), and change so that your API has RESTful naming (a noun in plural is a good rule of thumb)

HIPPIEKICK (changes requested):
Nice job Nathalie! Just remove the commented out code and update your readme with the new deploy 👍

EmOlsz (approved):

PR #120, opened 2020-12-13
Project Express API - Linda Hintze

lindahz:
 https://lindas-project-express-api.herokuapp.com/

It went well. My focus this week was to achieve only blue level requirements.

Nothing I can think of.


server.js
```diff
@@ -2,29 +2,80 @@ import express from 'express'
...
-// import netflixData from './data/netflix-titles.json'
-// import topMusicData from './data/top-music.json'
-
-// Defines the port the app will run on. Defaults to 8080, but can be 
-// overridden when starting the server. For example:
-//
-//   PORT=9000 npm start
+import harryData from './data/harry-potter-characters.json'
+
+const ERROR_CHARACTERS_NOT_FOUND = {error : 'No character results were found, please try again.'}
```
EmOlsz, line 7: Good idea to predefine error messages!

server.js
```diff
@@ -2,29 +2,80 @@ import express from 'express'
...
+// Starting route
+app.get('/', (request, response) => {
+  response.send('Hello, this is an API with Harry Potter characters. Go to https://lindas-project-express-api.herokuapp.com/characters to see all the characters. For more documentation, visit: https://github.com/lindahz/project-express-api')
+})
+
+// This endpoint shows all characters + will let you filter on name with query parameter
+app.get('/characters', (request, response) => {
+  const { name } = request.query
+  const filteredName = harryData.filter(character => 
+    character.name.toLocaleLowerCase() &&
```
EmOlsz, line 26: You don't really need that, this line of code is not doing anything. This is fine enough :

```
const filteredName = harryData.filter(character => character.name.toLocaleLowerCase().includes(name))
```

server.js
```diff
@@ -2,29 +2,80 @@ import express from 'express'
...
+
+// This endpoint shows all characters + will let you filter on name with query parameter
+app.get('/characters', (request, response) => {
+  const { name } = request.query
+  const filteredName = harryData.filter(character => 
+    character.name.toLocaleLowerCase() && 
+    character.name.toLocaleLowerCase().includes(name))
+
+  if (filteredName.length === 0) {
+    response.json(harryData)
```
EmOlsz, line 30: We shouldn't return list of all results if we didn't fulfill requirements of query.
If there were no results for our query (so, when `filteredName.length === 0`), we should return empty array. This is very important info for the user - "Aha, so I queried for character "Parry Hotter" and got empty array. That clearly means there is no book in database that has "Parry Hotter" character"

server.js
```diff
@@ -2,29 +2,80 @@ import express from 'express'
...
+  } else {
+    response.json(filteredName)
+  }
+})
+
+// This endpoint shows one character from a unique id
+app.get('/characters/id/:id', (request, response) => {
+  const { id } = request.params
+  const showId = harryData.find((item) => item.id === +id)
```
EmOlsz, line 39: I don't think `showId` is good name for variable, because you don't store any id in this variable. What you store there is either single character from Harry Potter universe or `undefined`

server.js
```diff
@@ -2,29 +2,80 @@ import express from 'express'
...
+  // Include error response here
+  if (!showId) {
+    response.status(404).json(ERROR_CHARACTERS_NOT_FOUND)
```
EmOlsz, line 42: Good job :)

server.js
```diff
@@ -2,29 +2,80 @@ import express from 'express'
...
+  } else {
+    response.json(showId)
+  }
+})
+
+// This enpoint shows all names of the characters, sorted from A-Z
+app.get('/characters/names', (request, response) => {
+  const names = harryData.map(name => name.name).sort()
```
EmOlsz, line 50: 👍

server.js
```diff
@@ -2,29 +2,80 @@ import express from 'express'
...
+  response.json(names)
+})
+
+// This endpoint shows all characters from a specific house
+app.get('/characters/:house', (request, response) => {
+  const { house } = request.params
+  const houses = harryData.filter((item) => item.house.toLocaleLowerCase() === house)
+
+  if (houses.length === 0) {
+    response.status(404).json(ERROR_CHARACTERS_NOT_FOUND)
```
EmOlsz, line 60: Once again, we should return empty array rather than error. Think about it like this - when we query for multiple elements, we should ALWAYS return array. It's either going to be array of elements (if we fulfilled queries) or just empty array.
We should return 404 error when we query for single element. Only when we queried for single element and that query was not fulfilled (for example looking for house "Technigo"), then we should return 404 error

EmOlsz (changes requested):

lindahz:
@EmOlsz could you please specify what changes are needed to meet blue level requirements? :)

server.js
EmOlsz: *** Needs to be fixed ***

server.js
EmOlsz: *** Needs to be fixed ***

server.js
EmOlsz: Regarding comment above, I suggest you to rename your variable to something like `character` or `queriedCharacter`

*** Needs to be fixed ***

server.js
EmOlsz: *** Needs to be fixed ***

EmOlsz:
Hello @lindahz , I've added new comment to each place in code that should be fixed :)
Look for comments : *** Needs to be fixed ***

lindahz:
Thanks Maks! Repo is now updated with the changes requested. 😊

EmOlsz:
Great, thank you, project approved :)

EmOlsz (approved):

PR #121, opened 2020-12-13
Ylva Nilsson - Project Express API

ylvan75:


server.js
```diff
@@ -1,33 +1,160 @@
...
-app.use(bodyParser.json())
+app.use(cors());
+app.use(bodyParser.json());
+
+const myEndpoints = require('express-list-endpoints');
 
-// Start defining your routes here
+// / endpoint (Root - Homepage)
+// RETURNS: A list of available endpoints (in an array)
 app.get('/', (req, res) => {
```
EmOlsz, line 22: Hmm this endpoint doesn't make much sense. I get what you wanted to achieve, but you should rather use `app.use()` method and combine it with `next()` method to detect if something went wrong. Moreover, 404 error should be used only if we didn't find records which fulfill our query (For example when we queried for a book with title "Technigo guides", but there was no book like this in database). In your case, I would rather suggest 500 or 503

server.js
```diff
@@ -1,33 +1,160 @@
...
+//     usage: /books/?title=harry
+app.get('/books', (req, res) => {
+  const { author, title } = req.query;
+  let booksList = booksData;
+  const totalNumberOfBooks = booksList.length;
+
+  // Query by author
+  if (author) {
+    booksList = booksList.filter((item) =>
+      item.authors.toLowerCase().includes(author.toLowerCase())
```
EmOlsz, line 50: 👍

server.js
```diff
@@ -1,33 +1,160 @@
...
+
+  // Query by title
+  if (title) {
+    booksList = booksList.filter((item) =>
+      item.title.toString().toLowerCase().includes(title.toLowerCase())
+    );
+  }
+
+  // PAGINATION by using slice - Default with 30 results per page
+  const page = req.query.page ? req.query.page - 1 : 0; // 0 as default
```
EmOlsz, line 62: Great job with pagination!

server.js
```diff
@@ -1,33 +1,160 @@
...
+    currentPage: page + 1,
+    pageSize: pageSize,
+    startIndex: startIndex,
+    endIndex: endIndex,
+    numberOfBooks: booksList.length,
+    results: booksListPerPage
+  };
+
+  // If the result is zero, status set to 404 and returning a useful data in the response
+  if (booksListPerPage.length === 0) {
```
EmOlsz, line 79: or just :

```
if (!booksListPerPage.length) { ... }
```

EmOlsz (approved):

PR #122, opened 2020-12-13
Sandrine's Spotify Releases API

MindstormingAB:
Link to my deployed project: https://spotify-releases.herokuapp.com/

My reflections on how this week's project turned out: Backend feels fun and fairly comprehensible so far, even if the choice between query and params is not obvious yet... It was quite challenging to dust off functions applicable to arrays and strings.

Things I'd like to have clarified or explained in more detail: --


server.js
```diff
@@ -1,33 +1,165 @@
...
+import spotifyData from "./data/spotify-releases.json";
+
+const port = process.env.PORT || 8080;
+const app = express();
+
+app.use(cors());
+app.use(bodyParser.json());
+
+// error message
+const ERROR_RELEASES_NOT_FOUND = { error: "No release matched your request" };
```
HIPPIEKICK, line 14: Nice!

server.js
```diff
@@ -1,33 +1,165 @@
...
+  };
+  if (filteredReleases.length === 0) {
+    res.status(404).json(ERROR_RELEASES_NOT_FOUND);
+  } else {
+    res.json({
+      total: filteredReleases.length,
+      releases: filteredReleases
+    });
+  }
+});
```
HIPPIEKICK, line 50: Way to go! 💯 Nice that you have many different queries and refilter the data. Nice with some metadata as well!

HIPPIEKICK: Case insensitivity would be like icing on the cake ;)

server.js
```diff
@@ -1,33 +1,165 @@
...
+
+// This route will return a single release based on id
+app.get("/releases/:id", (req, res) => {
+  const { id } = req.params;
+  const release = spotifyData.find(
+    (item) => item.id === id
+  );
+  if (!release) {
+    res.status(404).json(ERROR_RELEASES_NOT_FOUND);
+  } else {
```
HIPPIEKICK, line 60: Nice with some error handling as well!

server.js
```diff
@@ -1,33 +1,165 @@
...
+  );
+  if (filteredReleases.length === 0) {
+    res.status(404).json(ERROR_RELEASES_NOT_FOUND);
+  } else {
+    res.json({
+      total: filteredReleases.length,
+      releases: filteredReleases
+    });
+  };
+});
```
HIPPIEKICK, line 79: Endpoints should be named after what they return and not how they are filtered. This endpoint could be called `artists` (in plural) if that's what you're returning, but since you're returning releases, this endpoint is redundant - since you already have an `artist` query under the `/releases` endpoint :)

server.js
```diff
@@ -1,33 +1,165 @@
...
```
HIPPIEKICK, line 101: These could also be queries under your first endpoint:
`/releases?type=album&market=SE`

server.js
```diff
@@ -1,33 +1,165 @@
...
+
+// This route will return a list of all available release types.
+app.get("/types", (req, res) => {
+  const allTypes = spotifyData.map((item) => item.album_type);
+  const types = [...new Set(allTypes)].sort();
+  res.json({
+    total: types.length,
+    types
+  });
+});
```
HIPPIEKICK, line 111: This endpoint is named after what it's returning 👍

server.js
```diff
@@ -1,33 +1,165 @@
...
+
+// This route will return a list of all available released titles.
+app.get("/titles", (req, res) => {
+  const titles = spotifyData.map((item) => item.name.toString())
+    .sort((a, b) => a.toLowerCase().localeCompare(b.toLowerCase()));
+  res.json({
+    total: titles.length,
+    titles
+  });
+});
```
HIPPIEKICK, line 121: This too! Nice that you're trying out some different endpoints :)

server.js
```diff
@@ -1,33 +1,165 @@
...
+app.get("/markets", (req, res) => {
+  const allMarkets = spotifyData.map((item) => item.available_markets)
+    .join()
+    .split(",");
+  const markets = allMarkets.filter((item, index) => allMarkets.indexOf(item) === index).sort();
+  res.json({
+    total: markets.length,
+    markets
+  });
+});
```
HIPPIEKICK, line 147: Good naming 👍

server.js
```diff
@@ -1,33 +1,165 @@
...
+  res.send("Dummy endpoint that will return a list of all releases during the specified month");
+});
+
+app.get("/releases/start/:start/stop/:stop", (req, res) => {
+  res.send("Dummy endpoint that will return a list of all releases during the specified period");
+});
+
+app.get("/releases/theme/:theme", (req, res) => {
+  res.send("Dummy endpoint that will return a list of all releases with the specified theme");
+});
```
HIPPIEKICK, line 160: These could be queries as well ;)

HIPPIEKICK (approved):

PR #123, opened 2020-12-13
Collection of 10k books

BamStugendal:


server.js
```diff
@@ -1,30 +1,62 @@
...
+
+  if (topRated) {
+    booksFromYear = booksFromYear.filter((item) => item.average_rating > 4)
+  }
+
+  if (booksFromYear.length === 0) {
+    res.send('No books of that year exists in the collection')
+  }
+  res.json(booksFromYear)
+})
```
HIPPIEKICK, line 37: Endpoints should be named after what they return and not how they are filtered. These could therefore be query params under your first endpoint:
`/books?year=2010&toprated=true`

server.js
```diff
@@ -1,30 +1,62 @@
...
+app.get('/books/authors/:authors', (reg, res) => {
+  const authors = reg.params.authors
+  let booksByAuthors = booksData.filter((item) => item.authors === authors)
+
+  if (booksByAuthors.length === 0) {
+    res.send('No author of that name exists in the collection')
+  }
+  res.json(booksByAuthors)
+  //return only if the names are exactly the same.... 
+})
```
HIPPIEKICK, line 49: This too could be a query under your first endpoint since it's returning books but filtering on authors

server.js
```diff
@@ -1,30 +1,62 @@
...
+// Return books with given id
+app.get('/books/:id', (reg, res) => {
+  const id = reg.params.id
+  const booksById = booksData.find((item) => item.book_id === +id)
+
+  if (!booksById) {
+    res.send('No book of that id exists in the collection')
+  }
+  res.json(booksById)
 })
```
HIPPIEKICK, line 60: Nice and clean! Good with error handling

HIPPIEKICK (approved):

PR #124, opened 2020-12-13
Express-first-API

LillanEriksson:
Wish I had more time to work on this project. I only had the possibility to do the minimum requirements this week due to stressful times at work.

### DIFF
--- a/server.js
+++ b/server.js
@@ -5,11 +5,11 @@ import cors from 'cors';
 // If you're using one of our datasets, uncomment the appropriate import below
 // to get started!
 //
-import goldenGlobesData from './data/golden-globes.json';
+// import goldenGlobesData from './data/golden-globes.json';
 
 // import avocadoSalesData from './data/avocado-sales.json';
 // import booksData from './data/books.json'
-// import netflixData from './data/netflix-titles.json'
+import netflixData from './data/netflix-titles.json';
 // import topMusicData from './data/top-music.json'
 
 // Defines the port the app will run on. Defaults to 8080, but can be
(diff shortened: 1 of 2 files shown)


server.js
```diff
@@ -1,33 +1,71 @@
...
+	const showWin = req.query.win;
+	let nominationsFromYear = goldenGlobesData.filter(
+		(item) => item.year_award === +year
+	);
+
+	if (showWin) {
+		nominationsFromYear = nominationsFromYear.filter((item) => item.win);
+	}
+	res.json(nominationsFromYear);
+});
```
HIPPIEKICK, line 46: Endpoints should be named after what they return and not how they are filtered, so these could be query params under your `/nominations` endpoint:
`nominations?year=2010&win=true`

server.js
```diff
@@ -1,33 +1,71 @@
...
+
+//single item
+
+app.get('/year/:year/category/best_drama/?win=true', (req, res) => {
+	const year = req.params.year;
+	const showWin = req.query.win;
+
+	let bestDrama = goldenGlobesData.filter(
+		(item) => item.category === 'Best Motion Picture - Drama'
```
HIPPIEKICK, line 55: Would be nice if this was also a query param under the `nominations` endpoint 🤩 
`nominations?year=2010&win=true?category=drama`

HIPPIEKICK: Oh, this endpoint doesn't seem to be working 🤔 `?win=true` shouldn't be in the naming of the endpoint, but rather just a query param - right?

HIPPIEKICK (changes requested):
Fix the broken endpoint and you're good to go 😇

server.js
LillanEriksson: Changed the endpoint and it's working now

HIPPIEKICK: Looks like you removed it, or did you not commit your latest changes maybe. The requirements say to have one endpoint that only shows a single result

LillanEriksson: Fixed it now, changed dataset to have an ID to search for to show a single item

HIPPIEKICK (approved):

PR #127, opened 2020-12-13
Rebeka - API

rebkova:


server.js
```diff
@@ -22,9 +17,80 @@ const app = express()
 app.use(cors())
 app.use(bodyParser.json())
 
-// Start defining your routes here
-app.get('/', (req, res) => {
-  res.send('Hello world')
+// --- ALL THE ROUTES ---
+
+// --- front page ---
+app.get('/', (request /* incoming */, response /* outgoing */) => {
```
HIPPIEKICK, line 23: These inline comments make it look a little messy 🙈 Maybe keep this in your personal notes instead 😇

server.js
```diff
@@ -22,9 +17,80 @@ const app = express()
...
+// --- return ALL BOOKS & search by TITLE using QUERY SELECTORS ---
+//localhost:8080/books?title=freakonomics
+
+app.get("/books", (request, response) => {
+
+  const titleSearch = request.query.title
+
+  if (titleSearch === undefined) {
+    response.json(booksData)
+  }
```
HIPPIEKICK, line 38: Hmm, when would the titleSearch be `undefined`?

server.js
```diff
@@ -22,9 +17,80 @@ const app = express()
...
+})
+
+
+// --- getting SPECIFIC book based on ISBN13 ---
+//localhost:8080/id/:isbn13
+
+app.get("/id/:isbn13", (request, response) => {
+
+  const bookId = request.params.isbn13
```
HIPPIEKICK, line 60: Maybe call it isbn to avoid confusion 😇

server.js
```diff
@@ -22,9 +17,80 @@ const app = express()
...
+      response.status(404).json(`Ups, no keyword called '${titleLowerCase}' in a title.`)
+    }
+  }
+})
+
+
+// --- getting SPECIFIC book based on ISBN13 ---
+//localhost:8080/id/:isbn13
+
+app.get("/id/:isbn13", (request, response) => {
```
HIPPIEKICK, line 58: Endpoints should be named after what they return and not how they are filtered. This should therefore be named `/books/:isbn13`

server.js
```diff
@@ -22,9 +17,80 @@ const app = express()
...
+    else {
+      return 0
+    }
+  }
+
+  //sorting the numbers 
+  const sortedRatingArray = booksData.sort(compare)
+
+  response.json(sortedRatingArray)
 })
```
HIPPIEKICK, line 94: Wouldn't it be nice if this was a query under your main endpoint 🤩 
`/books?sort=descending`

HIPPIEKICK (approved):
Just think about RESTful naming (nouns in plural - name after what the endpoint return) :)

server.js
rebkova: I think if the title would have no entries i.e. the query would be empty. Like that: `/books?title `or `/books?title=` Correct me if I'm wrong.

server.js
rebkova: 👍

server.js
rebkova: 👍

PR #128, opened 2020-12-13
Week 17: Project express API

monkeybusiness101:


server.js
```diff
@@ -24,9 +13,27 @@ app.use(bodyParser.json())
 
 // Start defining your routes here
 app.get('/', (req, res) => {
-  res.send('Hello world')
+  res.send('We are advocates of avocados. /avocados for full list. /id/:id for specific id, /regions/:region for specific region')
```
HIPPIEKICK, line 16: Haha I love that you went for the avocados, think you're the first one 😄

server.js
```diff
@@ -24,9 +13,27 @@ app.use(bodyParser.json())
...
+
+app.get('/avocados', (req, res) => {
+  res.json(avocadoSalesData)
+})
+
+app.get('/avocados/id/:id', (request, response) => {
+  const { id } = request.params
+  const specificID = avocadoSalesData.find((avocado) => avocado.id === +id)
+  response.json(specificID)
 })
```
HIPPIEKICK, line 27: This could be called just `/avocados/:id` :)

server.js
```diff
@@ -24,9 +13,27 @@ app.use(bodyParser.json())
...
 
+app.get('/avocados/regions/:region', (request, response) => {
+  const { region } = request.params
+
+  let specificRegion = avocadoSalesData.filter((avocado) => avocado.region === region)
+  response.json(specificRegion)
+})
```
HIPPIEKICK, line 34: Nice with some filtering! Endpoints should be named after what they return and not how they are filtered so since you're returning avocados and not regions, this could instead be a query param under your main endpoint:
`/avocados?region=albany`

HIPPIEKICK (approved):

PR #129, opened 2020-12-13
Hanna's Book-API

Hanna-Skeppe:
It was a fun project this week. Still a bit confused as when best to use query-params and when to best use path-params (for finding books by specific author for example I included both query-param and path-param).
It was actually not so easy as I first thought to know how to structure the API, because the structure depends on the users needs.
It would be fun to continue further and learn more about PUT, POST, PATCH & DELETE. And also to build a front-end to the API to really understand the usability aspects more. I also commented in my code about the difference between using res.send and res.json and even though I read about the difference between them I would like to know more on when to use which of the two.
I completed the following levels/goals:
**🔵 Blue Level (Minimum Requirements)**
- Your API should have at least 2 routes. 
- A minimum of one endpoint to return a **collection** of results (array of elements)
- A minimum of one endpoint to return a **single** result (single element).
- Your API should be [RESTful](https://www.smashingmagazine.com/2018/01/understanding-using-rest-api/)
**🔴 Red Level (Intermediary Goals)**
- On routes which return a single item, handle when the item doesn't exist and return some useful data in the response.
- Accept filters via query parameters to filter the data you return from endpoints which return an array of data.
(- Create some empty/dummy endpoints which could contain more complex operations in the future.  Find good names for them (think back to the labs))Comment: I wrote down some dummy endpoints but I'm not sure they count as complex.
**⚫ Black Level (Advanced Goals)**
- If your dataset is large, try implementing 'pages' using `.slice()` to return only a selection of results from the array. 
(Create useful documentation for your endpoints. What's a good way to present this documentation?  What if it changes in the future?  Are there any npm packages that could help with this?) Comment: I didn't really get what this stretch-goal required. I installed 'express-list-endpoints' so that all available endpoints are shown in the root-path. But maybe this goal requires some other kind of documentation?


package.json
```diff
@@ -15,6 +15,7 @@
     "body-parser": "^1.19.0",
     "cors": "^2.8.5",
     "express": "^4.17.1",
+    "express-list-endpoints": "^5.0.0",
```
HIPPIEKICK, line 18: Yey, nice!

server.js
```diff
@@ -2,31 +2,182 @@ import express from 'express'
...
-app.get('/', (req, res) => {
-  res.send('Hello world')
+const myEndpoints = require('express-list-endpoints')
+// ERROR MESSAGES: 
+const ERROR_MESSAGE_AUTHOR_NOT_FOUND = { error: 'No book by that author found' }
+const ERROR_MESSAGE_BOOK_ISBN_NOT_FOUND = { error: 'No book with that ISBN-number found.' }
+const ERROR_MESSAGE_BOOK_ID_NOT_FOUND = { error: 'No book with that ID found.' }
+const ERROR_MESSAGE_LANGUAGE_NOT_FOUND = { error: 'No books with that language-code found' }
+const ERROR_MESSAGE_RATING_NOT_FOUND = { error: 'No books with that rating found' }
+const ERROR_MESSAGE_DATA_NOT_FOUND = { error: 'No data was found' }
```
HIPPIEKICK, line 28: Nice with constant error messages 💯

server.js
```diff
@@ -2,31 +2,182 @@ import express from 'express'
...
+// DEFINING ROUTES (endpoints):
+
+// ROOT-ENDPOINT.
+app.get('/', (req, res) => { // req is incoming(we don't change it), the res is outgoing(we do change it)
+  // NOTE TO SELF: Before the send we can write what ever we want here.
+  // For example:
+  // -Database connections
+  // -Complicated operations
+  // -Data lookup
+  // -Third party API-requests
```
HIPPIEKICK, line 39: Try to keep the comments shorter than the code 😇

server.js
```diff
@@ -2,31 +2,182 @@ import express from 'express'
...
+    ))
+  }
+
+  // Query by title:
+  if (title) {
+    booksList = booksList.filter(book => (
+      book.title.toString().includes(title)
+      || book.title.toString().toLowerCase().includes(title)
+    ))
+  }
```
HIPPIEKICK, line 70: Super nice with refiltering 👍

server.js
```diff
@@ -2,31 +2,182 @@ import express from 'express'
...
+    page: page,
+    firstPageIndex: firstPageIndex,
+    lastPageIndex: parseInt(booksList.length / pageSize),
+    results: bookPage
+  }
+
+  if (bookPage.length === 0) {
+    res.status(404).json(ERROR_MESSAGE_DATA_NOT_FOUND)
+  } else if (!page || !pageSize) {
+    res.json(booksList)
```
HIPPIEKICK, line 91: Given this code here, why do you have this
```
  const page = req.query.page ?? 0
  const pageSize = req.query.pageSize ?? 20
```
if not for default values? 😇

server.js
```diff
@@ -2,31 +2,182 @@ import express from 'express'
...
+// LANGUAGE-ENDPOINT. // According to Maks friday lecture: language should rather be query param??
+// Get books in a specified language (language_code):
+// Existing language-codes: 'eng', 'en-US', 'spa', 'fre', 'en-GB', 'mul', 'ger', 'ara', 'por', 'grc'
+app.get('/books/language/:language', (req, res) => {
+  const language = req.params.language
+  const booksByLanguage = booksData.filter((book) => book.language_code === language)
+  
+  // I used existingLanguages to console.log which unique languages that exists in the data.
+  // Would be nice to show to the user somehow which languages could be shown, but I didn't come that far.
+  // const existingLanguages = [...new Set(booksData.map(x => x.language_code))]
```
HIPPIEKICK, line 106: Yes, I would expect something like that given the name of the endpoint 😇 Endpoints should be named after what they return and not how they are filtered, so as it is now, this filtering could be a query under your main endpoint:
`/books?language=eng`

server.js
```diff
@@ -2,31 +2,182 @@ import express from 'express'
...
+  if (booksByLanguage.length > 0) {
+    res.send(booksByLanguage)
+  } else {
+    res.status(404).json(ERROR_MESSAGE_LANGUAGE_NOT_FOUND)
+  }
+})
+
+// ID-ENDPOINT. 
+// Get a single book by id:
+app.get('/books/id/:id', (req, res) => {
```
HIPPIEKICK, line 117: This could be called just `/books/:id`

server.js
```diff
@@ -2,31 +2,182 @@ import express from 'express'
...
+  const id = req.params.id
+  const singleBookId = booksData.find((book) => book.bookID === +id) // + turns string to number
```
HIPPIEKICK, line 119: Technically, this variable holds a whole bookObject, not just the id :)

server.js
```diff
@@ -2,31 +2,182 @@ import express from 'express'
...
+app.get('/books/ratings/:top100', (req, res) => {
+  const sortedBooks = booksData.sort((a, b) => b.average_rating - a.average_rating)
+  const top100Array = sortedBooks.slice(0, 100)
+
+  if (top100Array.length === 0) {
+    res.status(404).json(ERROR_MESSAGE_RATING_NOT_FOUND)
+  } else {
+    res.json(top100Array)
+  }
+})
```
HIPPIEKICK, line 155: This could also be a query under your main endpoint, something like:
`/books?sort=descending`

server.js
```diff
@@ -2,31 +2,182 @@ import express from 'express'
...
+      book.authors.includes(author) || book.authors.toLowerCase().includes(author)
+    ))
+
+    if (booksFilteredByAuthor.length > 0) {
+      res.json(booksFilteredByAuthor)
+    } else if (booksFilteredByAuthor.length === 0) {
+      res.status(404).json(ERROR_MESSAGE_AUTHOR_NOT_FOUND)
+    }
+  }
+})
```
HIPPIEKICK, line 174: This endpoint is redundant (you already have this filter as a query param ⭐ )

HIPPIEKICK (approved):

PR #133, opened 2020-12-13
project-express-api Marjaana

MayaBlixt:


src/server.js
```diff
@@ -0,0 +1,107 @@
...
+  //Calculate the end index
+  const endIndex = startIndex + +pageSize
+  const netflixItemsPerPage= netflixData.slice(startIndex, endIndex)
+  const returnObject = {
+    pageSize: pageSize,
+    page: page,
+    maxPages: parseInt(netflixData.length/ pageSize),
+    numItems: netflixItemsPerPage.length,
+    results: netflixItemsPerPage,
+  }
```
HIPPIEKICK, line 52: I think most of your endpoints could use some pagination 😇

src/server.js
```diff
@@ -0,0 +1,107 @@
...
+ * @apiSuccess {String} Netflix Movie and Tv show API .
+ *
+ */
+app.get('/', (req, res) => {
+  const myEndpoints = require('express-list-endpoints')
+  // Displays the endpoints that are available 
+  res.send(myEndpoints(app))
+})
+// Endpoint showing all tv shows and handles pageing returing 20 items
+app.get('/v1/netflixItems', (req, res) => {
```
HIPPIEKICK, line 36: I would go with lowercase

src/server.js
```diff
@@ -0,0 +1,107 @@
...
+// Endpoint showing all tv shows and handles query of country
+app.get ('/v1/TvShows', (req, res) => {
+  let tvShows = netflixData.filter((item) => item.type === "TV Show")
+  const selectedCountry = req.query.country
+  if (selectedCountry) {
+    tvShows = tvShows.filter((item) => item.country.includes(selectedCountry))
+
+  }
+  res.json(tvShows)
+})
```
HIPPIEKICK, line 90: Nice with some different filtering options! Wouldn't it be cool though if all these were queries under your main endpoint to give the user more freedom with the search options? Something like:
`/shows?type=movie&year=2010&country=sweden`

src/server.js
```diff
@@ -0,0 +1,107 @@
...
+app.get ('/v1/id/:id', (req, res) => {
+  const id = req.params.id
+  let idList = netflixData.filter((item) => item.show_id === +id)
+  // If id is not returing an item
+  if (idList.length === 0) { 
+    res.status(404).json(ERROR_MESSAGE_ID_NOT_FOUND)
+  } else{
+    res.json(idList)
+  }
+})
```
HIPPIEKICK, line 102: Since the id is supposed to be unique, it makes more sense to use `find()` than `filter()` here :) Super nice with error handling though! You can change `let` to `const` because you're not reassigning it. And remember RESTful naming, nouns in plural (named after what resource they return - so `/shows/:id` or similar)

HIPPIEKICK (approved):

PR #134, opened 2020-12-13
Express API Project - Sara Heijkenskjöld

sarahejhej:


KarolCislowski (approved):
This is a great project! 🌈 You are repeating a lot of code so in the future projects think about extracting some of the code to the separate functions to call it when needed 🤓

PR #139, opened 2020-12-15
express-api Anna Hasselström

annahasselstrom:
This was fun! I did not however push myself very far - instead I went back to React and did some repeating of conditional rendering. I needed that! 

I'm not at all sure when it comes to RESTful semantic and structure. Seems easy at first and then it isn't anymore 😅

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,13 +1,21 @@
-# Express API Project
+# Express API Project 📀
 
-Replace this readme with your own information about your project. 
+In this project I start my backend journey by creating an API using Express. The API have four RESTful endpoints which return an array of data, or a single item.
 
-Start by briefly describing the assignment in a sentence or two. Keep it short and to the point.
+## View it live 👀
 
-## The problem
+## The problem 🧠
 
-Describe how you approached to problem, and what tools and techniques you used to solve it. How did you plan? What technologies did you use? If you had more time, what would be next?
+- How to install and setup an Express server.
+- How to create routes.
+- How to deploy Node projects.
 
-## View it live
+## Learnings 🤓
 
-Every project should be deployed somewhere. Be sure to include the link to the deployed project so that the viewer can click around and see what it's all about.
+- The API endpoints I created use array methods such as .find(), and .filter().
+- It has four routes. 
+- Endpoints to return a **collection** of results (array of elements)
+- Endpoints to return a **single** result (single element).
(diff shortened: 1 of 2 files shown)


server.js
```diff
@@ -1,33 +1,96 @@
...
+  const showWon = request.query.win
+  console.log(showWon)
+
+  let nominationsFromYear = goldenGlobesData.filter((item) => item.year_award === +year);
+  
+  if (showWon) {
+    nominationsFromYear = nominationsFromYear.filter((item) => item.win)
+  } 
+  response.json(nominationsFromYear)
+});
```
KarolCislowski, line 72: those endpoints are very nice 👍🏻

server.js
```diff
@@ -1,33 +1,96 @@
...
+  response.send('Hello world again')
+});
+
+// Define the method - GET. Give it a path - 'nominations'. Handle the response: get 
+// all data and send as a json response to client.
+// The url: /nominations
+app.get('/nominations', (request, response) => {
+  response.json(goldenGlobesData)
+});
+
```
KarolCislowski, line 32: You could add the `/nominations/:index` and return one object the easy way by using array's indexes like id from the database 😺

server.js
```diff
@@ -1,33 +1,96 @@
...
+  let bestCategoriesFromYear = goldenGlobesData.filter(item => {
+    return (
+      item.category === category && item.year_award === +year
+    )
+  });
+
+// Tried < 0 but got an empty array back. Cause filter returns an array.
+  if (bestCategoriesFromYear < 1) {
+    response.status(404).json(ERROR_NOT_FOUND);
```
KarolCislowski, line 52: It is great that you provided the error messages 👍🏻

README.md
```diff
@@ -1,13 +1,23 @@
...
+
+- The API endpoints I created use array methods such as .find(), and .filter().
+- It has four routes. 
+- Endpoints to return a **collection** of results (array of elements)
+- Endpoints to return a **single** result (single element).
+- The API is [RESTful](https://www.smashingmagazine.com/2018/01/understanding-using-rest-api/)
 
-## View it live
 
-Every project should be deployed somewhere. Be sure to include the link to the deployed project so that the viewer can click around and see what it's all about.
```
KarolCislowski, line 13: It would be nice to have some instructions on what endpoints you are providing and how to use them 😺

KarolCislowski (approved):
This is a nice and ambitious project 😸 Great work!

PR #144, opened 2021-05-07
Express API - Johan Löfgren

johlofg:
https://johan-music-api.herokuapp.com/

### DIFF
--- a/server.js
+++ b/server.js
@@ -1,4 +1,4 @@
-import express, { request } from 'express'
+import express from 'express'
 import bodyParser from 'body-parser'
 import cors from 'cors'
 import listEndpoints from 'express-list-endpoints'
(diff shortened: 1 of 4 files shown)


server.js
```diff
@@ -1,34 +1,64 @@
...
+// Endpoint to get songs from specific artist
+app.get('/artist/', (req, res) => {
+  const { artistName } = req.query 
+  if (artistName) {
+    // eslint-disable-next-line max-len
+    const artistList = topMusicData.filter((item) => item.artistName.toLowerCase().includes(artistName))
+    res.json(artistList)
+  }
+  res.json(topMusicData)
+})
```
KarolCislowski, line 42: when you want to use a query to filter your data you don't create an extra endpoint. you should add this logic to the `/songs` endpoint.

server.js
```diff
@@ -1,34 +1,64 @@
...
+
+// Endpoint to get one song
+app.get('/song/:id', (req, res) => {
```
KarolCislowski, line 45: endpoints should be in plural form so: `/songs/:id`

KarolCislowski (approved):
nice API 🦄

PR #158, opened 2021-05-08
express-api - Ati Takrami

takrami:
https://golden-globe-awards-api.herokuapp.com


README.md
```diff
@@ -1,13 +1,7 @@
...
-
-## The problem
-
-Describe how you approached to problem, and what tools and techniques you used to solve it. How did you plan? What technologies did you use? If you had more time, what would be next?
+This is a simple golden globes API.
 
 ## View it live
 
-Every project should be deployed somewhere. Be sure to include the link to the deployed project so that the viewer can click around and see what it's all about.
+https://golden-globe-awards-api.herokuapp.com/
```
KarolCislowski, line 7: It would be very nice to have some instructions, how to use your API. Remember to make documentation on what endpoints and queries your user can use. 🤓

server.js
```diff
@@ -1,34 +1,67 @@
...
-// import goldenGlobesData from './data/golden-globes.json'
-// import avocadoSalesData from './data/avocado-sales.json'
-// import booksData from './data/books.json'
-// import netflixData from './data/netflix-titles.json'
-// import topMusicData from './data/top-music.json'
-
-// Defines the port the app will run on. Defaults to 8080, but can be 
-// overridden when starting the server. For example:
-//
-//   PORT=9000 npm start
```
KarolCislowski, line 17: nice cleaning 🧹

server.js
```diff
@@ -1,34 +1,67 @@
...
+
+  const ceremoniesUnique = [];
+  ceremoniesDuplicated.forEach((item) => {
+    if (!ceremoniesUnique.includes(item)) {
+      ceremoniesUnique.push(item);
+    }
+  });
+
+  res.json({ data: ceremoniesUnique });
+});
```
KarolCislowski, line 27: returns `Internal Server Error` with status 500 😭

server.js
```diff
@@ -1,34 +1,67 @@
...
+    );
+  }
+
+  if (win) {
+    result = result.filter((movie) => movie.win);
+  }
+
+  if (category) {
+    result = result.filter((movie) => movie.category.includes(category));
+  }
```
KarolCislowski, line 45: nice filtering 💯

server.js
```diff
@@ -1,34 +1,67 @@
...
+
+  if (filteredFilm.length > 0) {
+    response.json(filteredFilm);
+  } else {
+    response
+      .status(404)
+      .json({ message: `There is not information about ${film}` });
+  }
+});
+
```
KarolCislowski, line 64: To meet blue requirments you need to provide:
- A minimum of one endpoint to return a **single** result (single element).

KarolCislowski (changes requested):
Please provide an endpoint that is returning a single element.

server.js
takrami: This will be returned one film. Actually, the golden globe data doesn't have any other unique value.

KarolCislowski: It is not a single result (you have 3 records here) You need to provide an endpoint which returns always 1 record. image

takrami: Thanks for putting time into this :)
Let me explain it a bit more about why my API cannot return unique result. So as you can see the data structure provided by Technigo:
```
{
    "year_film": 2009,
    "year_award": 2010,
    "ceremony": 67,
    "category": "Best Performance by an Actress in a Motion Picture - Musical or Comedy",
    "nominee": "Meryl Streep",
    "film": "Julie & Julia",
    "win": true
}
```
There is no field that can be unique because there is no such a `filmId` here.
`year_film` => Many films can be produced in a year.
`year_award` => Many films can be awarded in a year.
`ceremony` => It's very clear that it cannot be a unique field.
`category` => Many films can be under a category. We have many comedy films.
`nominee` => I think your image can explain that it cannot be unique.
`film` => Is the film name and based on the image you sent it cannot be unique.
`win` => It's clear that it cannot be unique.
Based on what I explained. I thought that the most probable unique field can be `film`. But I can see that is not.
I believe the data structure is not proper enough to have a single endpoint although you have any suggestions about the unique field I would be really happy to implement that :)
Btw if there was an `id` property the code would be like this:
```js
app.get("/golden-globes/:id", (request, response) => {
  const { id } = request.params;
  const foundFilm = goldenGlobesData.find((movie) => {
    return movie.id === id;
  });

  if (foundFilm) {
    response.json(foundFilm);
  } else {
    response
      .status(404)
      .json({ message: `No film has found with this id: ${id}` });
  }
});
```

KarolCislowski (approved):
I can see that you understand the topic 👍🏻  You could use the index as the ID but I would approve it now 🥇

takrami:
> I can see that you understand the topic 👍🏻 You could use the index as the ID but I would approve it now 1st_place_medal

Thank you so much for approval :) :+1:  
I think using index as ID would be problematic. Maybe not here because the data is not gonna change and it's sort of static in the database. Otherwise, in the case of row deletion, all indexes after that row would change and this would defeat the whole purpose of the uniqueness of ID.
Thanks again :100:

PR #182, opened 2021-05-09
Project Express-API Maria Sparre 

MarySparrow:
https://netflixmovies.herokuapp.com/movies


server.js
```diff
@@ -1,34 +1,83 @@
...
+// Start route
+app.get('/', (request, response) => {
+  response.send(listEndpoints(app))
+})
+// return all movies 
+app.get('/movies', (request, response) => {
+  response.json(movies)
+})
+//return one single movie based on known ID
+app.get('/id/:id', (request, response) => {
```
KarolCislowski, line 25: REST naming convention is to provide the name of the list you are filtering, so here you should use `/movies/:id` 
When you filtering by the primary key (id here) you do not specify the key, so `/movies/id/:id` would be wrong as well.

server.js
```diff
@@ -1,34 +1,83 @@
...
+  const { id } = request.params 
+  const filteredMovies = movies
+    .find(movie => movie.show_id === +id
+  )
+filteredMovies ? response.json(filteredMovies) : response.status(404).send(`There is no movie with ID "${id}" in the system`)
+})
+//return movie based on year of release
+app.get('/year/:year', (request, response) => {
```
KarolCislowski, line 33: Again you are filtering movies so it should be `/movies/years/:year`

server.js
```diff
@@ -1,34 +1,83 @@
...
+    item => item.release_year === +year
+  )
+  releaseYear.length !== 0
+    ?  
+      response.json({ length: releaseYear.length, data: releaseYear}) //json vs send?
+    :  
+      response.send({ length: releaseYear.length, data: releaseYear})
 })
+//return type of movies, filtered out to not show duplicates
+app.get('/type/type', (request, response) => {
```
KarolCislowski, line 45: and remember that you need to use plural forms in your endpoints `/movies/types/:type`

server.js
```diff
@@ -1,34 +1,83 @@
...
+  }
+  if(country) {
+    data = data 
+      .filter(item => item.country.toLowerCase().includes(country.toLowerCase()))
+  }
+  if(title){
+    data = data
+      .filter(item => item.title.toString().toLowerCase().includes(title.toLowerCase()))
+  }
+  response.json({ length: data.length, data: data})
```
KarolCislowski, line 72: nice filtering

KarolCislowski (changes requested):
Please change your endpoints to meet REST convention 🤓

MarySparrow:
Hello Karol,

Thanks for clarifying REST convention :)

One part I am not sure if I have mistaken still, (see snippet below)  you
write /movies/types/:type but will it also work typing /movies/types/type
with the filtering I have set up ?

Little confused in the learning here still.

Thanks!
Maria

server.js
<https://github.com/Technigo/project-express-api/pull/182#discussion_r633072889>
:

>  })

+//return type of movies, filtered out to not show duplicates

+app.get('/type/type', (request, response) => {


and remember that you need to use plural forms in your endpoints
/movies/types/:type
------------------------------


Den sön 16 maj 2021 kl 12:20 skrev Karol Cislowski ***@***.***

KarolCislowski (approved):
Nice app 👍🏻

PR #184, opened 2021-05-09
Project-express-api Marielle Ohlsson

Marielle8:
I aimed for the blue requirments this week but I don't think the project works like it should. I got a lot of errors and couldn't open up the app in heroku before I handled it in.

### DIFF
--- a/server.js
+++ b/server.js
@@ -35,7 +35,7 @@ app.use(bodyParser.json())
 //})
 
 app.get('/books', (request, response) => {
-  res.json(booksData)
+  response.json(booksData)
 })
 
 //Endpoint to get one book
--- a/server.js
+++ b/server.js
@@ -1,6 +1,7 @@
 import express, { request, response } from 'express'
 import bodyParser from 'body-parser'
 import cors from 'cors'
+import listEndpoints from 'express-list-endpoints'
 
 import booksData from './data/books.json'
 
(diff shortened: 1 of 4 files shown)


server.js
```diff
@@ -1,34 +1,49 @@
...
-// import goldenGlobesData from './data/golden-globes.json'
-// import avocadoSalesData from './data/avocado-sales.json'
-// import booksData from './data/books.json'
-// import netflixData from './data/netflix-titles.json'
-// import topMusicData from './data/top-music.json'
-
-// Defines the port the app will run on. Defaults to 8080, but can be 
-// overridden when starting the server. For example:
-//
-//   PORT=9000 npm start
```
KarolCislowski, line 17: Nice that you cleaned your code 👍🏻

server.js
```diff
@@ -1,34 +1,49 @@
...
+  const { bookid } = request.params
+  const book = booksData.find(book => book.bookID === +bookid)
+  if (!book) {
+    response.status(404).send(`Sorry, book ${bookid} dosen't exist`)
+  }
+  response.json(book)
+})
+
+//Endpoint to get a specifik author/authors
+app.get('/author', (request, response) => {
```
KarolCislowski, line 38: your endpoint should be always in the plural form.

server.js
```diff
@@ -1,34 +1,49 @@
...
+  const { authors } = request.query
+  if (authors) {
+    const authorsList = booksData.filter(author => author.authors.toLowerCase().includes(authors))
```
KarolCislowski, line 41: You are filtering books by authors here, not the author list, so it should be the part of `/books` endpoint

KarolCislowski (approved):
Nice api 👍🏻

PR #212, opened 2021-12-10
Week 17, Maria Thomasson

MariaThomasson:
https://week17.herokuapp.com/

### DIFF
--- a/server.js
+++ b/server.js
@@ -1,19 +1,9 @@
 import express from 'express'
 import cors from 'cors'
+import listEndpoints from 'express-list-endpoints'
 
-// If you're using one of our datasets, uncomment the appropriate import below
-// to get started!
-// 
-// import goldenGlobesData from './data/golden-globes.json'
-// import avocadoSalesData from './data/avocado-sales.json'
-// import booksData from './data/books.json'
-// import netflixData from './data/netflix-titles.json'
-// import topMusicData from './data/top-music.json'
+import booksData from './data/books.json'
 
-// Defines the port the app will run on. Defaults to 8080, but can be 
-// overridden when starting the server. For example:
-//
-//   PORT=9000 npm start
 const port = process.env.PORT || 8080
 const app = express()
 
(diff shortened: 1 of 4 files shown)


server.js
```diff
@@ -23,7 +13,72 @@ app.use(express.json())
...
+// Endpoint for all books:
+// search query title and author:
+app.get('/books', (req, res) => {
+  const { title, author } = req.query
+  
+  let books = booksData.sort((a, b) => a.bookID - b.bookID)
+
+  if (title) {
+    // checks if book title is a string (because some are not!)
+    books = books.filter((book) => typeof book.title === "string")
```
nehrwein, line 36: what other value-type of do you get from the dataset?

server.js
```diff
@@ -23,7 +13,72 @@ app.use(express.json())
...
+    // checks for title
+    books = books.filter((book) => book.title.toLowerCase().includes(title.toLowerCase()))
+  }
+
```
nehrwein, line 40: nice, that you use the includes method to filter the books and author!

server.js
```diff
@@ -23,7 +13,72 @@ app.use(express.json())
...
+    // displays 404 error message if author or title search has no results
+    res.status(404).json('Not Found')
+  } else {
+    res.json(booksData)
+  }
+})
+
+// Endpoint that shows all authors in alphabetic order:
+app.get('/books/authors/:authors/', (req, res) => {
+  const listAuthors = booksData.map((item) => item.authors).sort()
```
nehrwein, line 56: `const listAuthors = Array.from(new Set(booksData.map((item) => item.authors))).sort()`
This will create a list of all authors in the Books-Data without any duplicates. It's creating a Set (object with unique items). Afterwards you convert the set into an array and sort it alphabetically.
I can explain this further, if you want :)

nehrwein (commented):
Hej Maria,
congrats to your first Backend-Project. Everything works as it should and your code looks clean and pretty straight forward. I like that you use query params and have implemented lots of endpoints to choose from! I left some comments in the code, that might be helpful for further improvement of the code. 🙂 
Well done! 👍 🥇 
Birgit

PR #214, opened 2021-12-10
Johanna MJ Project API

JohannaMJ:
This is a backend project building an API in NODE using Express. The API has a bunch of routes to get to endpoints. The data has been manipulated by filtering and limiting arrays so that different data is displayed depending on the endpoint.

I started the project by choosing a JSON file to work with (Netflix shows) and to create a first endpoint that displays all objects in the JSON file. I then tried out params to create an endpoint that displays a single object from the JSON file, using the find-method to catch specific object ids. I continued creating more endpoints with params, and then with queries, using the filter method. By using queries I created endpoints to display arrays of objects depending on the title, type of show (tv-series or movie), country, cast and release year. The queries can be combined so that you can get to an endpoint that returns an array of object from a specific country and whose title includes a specific name.
If there are no objects to show but the response from the JSON was successful, it returns an empty array.
If I had more time I would create a frontend webpage to fetch and display the data from the API.

https://flix-data.herokuapp.com/


server.js
```diff
@@ -1,33 +1,104 @@
...
+
 app.get('/', (req, res) => {
-  res.send('Hello world')
-})
+  res.json(listEndpoints(app));
+});
+
+//endpoints by using one or more queries
+app.get('/shows', (req, res) => {
+  const { title, type, cast, country, year } = req.query;
```
madeleinesvensson, line 19: Really cool that you made a endpoint than can take so many queries :)

server.js
```diff
@@ -1,33 +1,104 @@
...
+    show.listed_in.toString().toLowerCase().includes(category.toLowerCase())
+  );
+
+  res.json({
+    response: showByCategory,
+    success: true,
+  });
+});
+
+// endpoint to get all shows from a specific year, and also possible to add query to sort tvshows or movies
```
madeleinesvensson, line 82: Comments, thank you so that I understand what they do 👍

server.js
```diff
@@ -1,33 +1,104 @@
...
+  });
+});
+
+// endpoint to get a specific shows by id
+app.get('/shows/id/:id', (req, res) => {
+  const { id } = req.params;
+  const showID = netflixData.find((show) => show.show_id === +id);
+
+  if (!showID) {
+    res.status(404).send(`No show found with id number ${id} :(`);
```
madeleinesvensson, line 59: Nice to include the id in the error message

madeleinesvensson (commented):
Very nicely done! Really cool that you tried out so many different endpoints and also added all the different queries to filter on in the /shows endpoint. The code file is clean and easy to follow along to and the comments are so helpful to understand what each endpoint leads to and shows. The api works and gets the data correct as well! :D I don't have anything to point out as an improvement so good job! Keep it up!!

PR #217, opened 2021-12-10
express-api-birgit

nehrwein:
Visit my project: https://nehrwein-netflix-api.herokuapp.com/

### DIFF
--- a/server.js
+++ b/server.js
@@ -7,7 +7,7 @@ import cors from 'cors'
 // import goldenGlobesData from './data/golden-globes.json'
 // import avocadoSalesData from './data/avocado-sales.json'
 // import booksData from './data/books.json'
-// import netflixData from './data/netflix-titles.json'
+import netflixData from './data/netflix-titles.json'
 // import topMusicData from './data/top-music.json'
 
 // Defines the port the app will run on. Defaults to 8080, but can be 
@@ -21,9 +21,22 @@ const app = express()
 app.use(cors())
 app.use(express.json())
 
-// Start defining your routes here
+// Start defining routes here
 app.get('/', (req, res) => {
-  res.send('Hello world')
+  res.send('Welcome to my API. You can find some selected Netlify-data here')
+})
+
+// route provides a sorted list of all countries that are in netflixData
+app.get('/countries', (req, res) => {
+  const countries = Array.from(new Set(netflixData.map(item => item.country))).sort()
+  res.json(countries)
+})
+
+// route with all movies from the provided country
+app.get('/countries/:country', (req, res) => {
+  const { country } = req.params
+  const moviesByCountry = netflixData.filter((item => item.country.toLowerCase() === country))
+  res.json(moviesByCountry)
 })
 
 // Start the server
--- a/server.js
+++ b/server.js
@@ -3,7 +3,6 @@ import cors from 'cors'
 import listEndpoints from 'express-list-endpoints'
 import netflixData from './data/netflix-titles.json'
 
-
 // Defines the port the app will run on. Defaults to 8080, but can be 
 // overridden when starting the server. For example:
 //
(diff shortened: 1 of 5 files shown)


server.js
```diff
@@ -21,13 +14,135 @@ const app = express()
...
+app.get('/endpoints', (req, res) => {
+  res.json({
+    response: listEndpoints(app),
+    success: true
+  })
+})
+
+// route provides a sorted list of all countries that are in netflixData
+app.get('/countries', (req, res) => {
+  const countries = Array.from(new Set(netflixData.map(item => item.country))).sort()
```
DALA746, line 35: Could you leave a comment about this line of code? :)

server.js
```diff
@@ -21,13 +14,135 @@ const app = express()
...
+  res.json({
+    response: filteredMovies,
+    success: true
+  })
+})
+
+// route provides one movie by ID
+app.get('/movies/id/:id', (req, res) => {
+  const { id } = req.params
+  const movies = netflixData.filter(item => item.type === "Movie")
```
DALA746, line 106: Here you are filtering by type, maybe rename this variable to something else? Maybe filteredByType?

server.js
```diff
@@ -21,13 +14,135 @@ const app = express()
...
+  }  
+
+  res.json({
+    response: filteredMovies,
+    success: true
+  })
+})
+
+// route provides one movie by ID
+app.get('/movies/id/:id', (req, res) => {
```
DALA746, line 104: As I understand, this route gives only a movie with a specific id, but why do you filter on line 106? When using params it only returns one specific item as I understand 🤔

DALA746: Aaa, now I saw that you are using movies variable on line 108 :) maybe it is a good idea to change variable name :)

server.js
```diff
@@ -21,13 +14,135 @@ const app = express()
...
+      success: true
+    })
+  } 
+})
+
+// route provides all movies and has the possibility to query for director, year and actor. You can also do pagination by setting page & limit as query parameters
+app.get('/movies', (req, res) => {
+  const { director, year, actor, page, limit } = req.query
+
+  const movies = netflixData.filter(item => item.type === "Movie")
```
DALA746, line 68: You only have movies, not TV shows in your project so that is why you are filtering by type :)

DALA746 (commented):
Readable and understandable code to read! Nice that you implemented documentation with Postman! I did my project with the same data so I couldn't add more feedback! You implemented a different approach on line 35 which is great, but leave the comment next time just to explain the code. Nice job! 💯

server.js
nehrwein: Sure :)
The aim with this code is to show a list of all countries in the Netflix-Data without any duplicates. Therefore I am creating a Set (object with unique items). I convert the set into an array afterwards and sort it alphabetically

server.js
nehrwein: now I made movies a global variable. I hope that's more clear now and also has the advantage of less repetitive code. Thanks for pointing out!

server.js
```diff
@@ -1,33 +1,147 @@
...
+  res.json({
+    response: listEndpoints(app),
+    success: true
+  })
+})
+
+// route provides a sorted list of all countries that are in netflixData
+app.get('/countries', (req, res) => {
+  // a list of all countries in the Netflix-Data without any duplicates. Therefore I am creating a Set (object with unique items). I convert the set into an array afterwards and sort it alphabetically
+  const countries = Array.from(new Set(netflixData.map((item) => item.country))).sort()
```
Jsfrulle, line 38: oh good code :D this gives the coutries in alphabetical order i guess

server.js
```diff
@@ -1,33 +1,147 @@
...
+    })
+  } else {
+    res.json({
+      response: movie,
+      success: true
+    })
+  }
+})
+
+// route provides movies by name (can return more than one movie, if the provided parts of the title match with several movies)
```
Jsfrulle, line 123: really nice way to let the frontender use this array!!! Like if you want to find all film with a word in its title!!!! this is a setup up compare to what i done in my project!

server.js
Jsfrulle: maybye add a 404 respons

PR #240, opened 2021-12-12
Express API w17 Lisa Pousette Blomé

annaester:


server.js
```diff
@@ -1,33 +1,132 @@
...
 // to get started!
-// 
+//
 // import goldenGlobesData from './data/golden-globes.json'
 // import avocadoSalesData from './data/avocado-sales.json'
-// import booksData from './data/books.json'
+
+import booksData from "./data/books.json";
+
+// console.log(booksData.length); -- 450st!
```
Skrosen, line 12: Remember to delete console.logs (when you don't need them ;) )

server.js
```diff
@@ -1,33 +1,132 @@
...
+
+  // if (startIndex > 0) {
+  //   results.previous = {
+  //     page: page - 1,
+  //     limit: limit,
+  //   };
+  // }
+
+  // results.results = booksData.slice(startIndex, endIndex);
+  // res.send(results);
```
Skrosen, line 63: I realize you probably want to continue working on this but if you are not planing on doing it quite soon it can be very easy to forget about it and then this out commented code will take up space/make the code less clean :)

server.js
```diff
@@ -1,33 +1,132 @@
...
+});
+
+//get a specific book based on id, using param
+app.get("/books/bookid/:bookID", (req, res) => {
+  const { bookID } = req.params;
+
+  const bookDetail = booksData.find((book) => book.bookID === +bookID);
+
+  if (!bookDetail) {
+    res.status(404).send("no book found");
```
Skrosen, line 103: I like that you send the status code together with the response to clarify it for frontend.

server.js
```diff
@@ -1,33 +1,132 @@
...
+    res.json(bookDetail);
+  }
+  //where you can get the title, author, rating, language for each book in your frontend
+});
+
+//all the books with top rating (at least 4/5)
+app.get("/books/topbooks", (req, res) => {
+  const topRatedBooks = booksData.filter((book) => book.average_rating >= +4);
+  res.json(topRatedBooks);
+});
```
Skrosen, line 114: Nice endpoint!

server.js
```diff
@@ -1,33 +1,132 @@
...
+app.get("/books/authors/:authors", (req, res) => {
+  const allAuthors = booksData.map((item) => item.authors).sort();
+  res.send(allAuthors);
+});
+
+//just at test to post in postman
+app.post("/postbook", (req, res) => {
+  const { body } = req;
+  res.send(body);
+});
```
Skrosen, line 126: Cool that you tried out the post function here!

server.js
```diff
@@ -1,33 +1,132 @@
...
-app.use(express.json())
+app.use(cors());
+app.use(express.json());
 
 // Start defining your routes here
-app.get('/', (req, res) => {
-  res.send('Hello world')
-})
+app.get("/", (req, res) => {
+  res.send("Please use an endpoint to fetch requests!");
```
Skrosen, line 30: A tip is to add express-list-endpoints here so that the user can see all the endpoints from start, or documentation as we have talked about later on. :)

Skrosen (commented):
Really nice project and you are very good with commenting your code and making it easy for the reader to understand what a specific part of the code is doing. Something that I thought about (I will think out loud here so if you have heard anything about best practice you are very welcome to write to me as I'm looking in to it more :)) As I read (se below) the endpoints should be in order from least specific to most specific, however, you are not using the exact same route to your endpoints as they do in the example below, so in this case it's nothing to worry about. But I was thinking if there is a best practice, either it could be putting all endpoints with :something in the bottom to avoid missing something further on when you're working with a lot of endpoints. Or is it clustering them in the code in some way like, all books/autors in one place and all books/titles in one and so on. Do you know?
I read about it here: https://www.notion.so/Week-17-Build-an-API-using-Express-9a3409a2a4ef49c7908bf529e61c2122

PR #242, opened 2021-12-12
elsa bjelkedal

Elsa-Johanna:
https://technigo-backend-elsa-b.herokuapp.com/


the-j-curl (changes requested):
This is a really nice API Elsa.
I have left a couple of comments to improve your code and align with the Techinigo code guidelines. Then it will be complete . Great work!

PR #245, opened 2021-12-12
Sofia - the foxes - express API

efstasia:


server.js
```diff
@@ -1,33 +1,83 @@
...
+    );
+  }
+
+  if (theOfficeDataToSend.length === 0) {
+    res
+      .status(404)
+      .json(
+        "Sorry we couldn't find what you were looking for or it doesn't exist"
+      );
+  } else {
```
Ajliin, line 75: Nice and clean! The only thing I could think about is do a status for 400 as well as approvment for next time :)

Ajliin (commented):
Really nice work and met all requirements!

PR #249, opened 2021-12-13
BooksAPI-Naushin Malik

nama0027:
## View it live

https://booksapi-naushinmalik.herokuapp.com/

### DIFF
--- a/server.js
+++ b/server.js
@@ -1,33 +1,111 @@
-import express from 'express'
-import cors from 'cors'
+import express from 'express';
+import cors from 'cors';
+import listEndPoints from 'express-list-endpoints';
 
-// If you're using one of our datasets, uncomment the appropriate import below
-// to get started!
-// 
 // import goldenGlobesData from './data/golden-globes.json'
 // import avocadoSalesData from './data/avocado-sales.json'
-// import booksData from './data/books.json'
+import booksData from './data/books.json';
 // import netflixData from './data/netflix-titles.json'
 // import topMusicData from './data/top-music.json'
 
-// Defines the port the app will run on. Defaults to 8080, but can be 
+// Defines the port the app will run on. Defaults to 8080, but can be
 // overridden when starting the server. For example:
 //
 //   PORT=9000 npm start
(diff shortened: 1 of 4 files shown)


server.js
```diff
@@ -1,33 +1,235 @@
...
-app.use(cors())
-app.use(express.json())
+app.use(cors());
+app.use(express.json());
+
+// middleware for pagination and other filtration based on search criteria
+
+function filteredAndPaginatedResults(booksDataToSend) {
+  return (req, res, next) => {
+    const {
```
Fatima-GR, line 23: Good to see that you used so many queries in one single endpoint

server.js
```diff
@@ -1,33 +1,235 @@
...
+      filteredBooksData = filteredBooksData.filter(
+        (book) => book.num_pages >= +minPages
+      );
+    }
+    if (maxPages) {
+      filteredBooksData = filteredBooksData.filter(
+        (book) => book.num_pages <= +maxPages
+      );
+    }
+    //* --------------- for pagination -------- *//
```
Fatima-GR, line 81: Like always, you went an extra mile by doing the pagination, well done!

server.js
```diff
@@ -1,33 +1,235 @@
...
+// route for all books or books based on a specific search criteria
+app.get('/books', filteredAndPaginatedResults(booksData), (req, res) => {
+  res.json({
+    response: res.paginatedResults,
+    success: true,
+  });
+});
+
+// route for the books having  the specified rating
+app.get('/books/rating/:rating', (req, res) => {
```
Fatima-GR, line 133: It is recommended when using path params, to do an if statement so if the data does not arrive or there is a problem with it, the user or frontend developers can get a message and also the 404 status to know what is going on!
Also, when using path params, the correct method to use is the find() method, since you are expecting to get only one single element. On the other hand, the filter method is used for the query params because you expect more than one element that passes the test.

Fatima-GR (commented):
Hi Naushin! Great job! Your code is very well documented, organized and more advanced than the code that the coaches provided us!

Fatima

PR #252, opened 2021-12-14
project-express-api-by HelenaW

HelenaW86:


README.md
```diff
@@ -1,13 +1,6 @@
 # Express API Project
 
-Replace this readme with your own information about your project. 
-
-Start by briefly describing the assignment in a sentence or two. Keep it short and to the point.
-
-## The problem
-
-Describe how you approached to problem, and what tools and techniques you used to solve it. How did you plan? What technologies did you use? If you had more time, what would be next?
```
JuliaNiki, line 9: I would recommend you to write some short description for this project so it would be clear for anyone looking on your Github, what is this project about :) Take some inspo from week description in Notion

server.js
```diff
@@ -1,33 +1,146 @@
...
-//
-//   PORT=9000 npm start
-const port = process.env.PORT || 8080
-const app = express()
+import express from "express";
+import cors from "cors";
+import expressListEndpoints from "express-list-endpoints";
+
+import data from "./data/golden-globes.json";
+import books from "./data/books.json";
```
JuliaNiki, line 6: Great idea to play with both of these datasets and make endpoints for both!

server.js
```diff
@@ -1,33 +1,146 @@
...
+
+const port = process.env.PORT || 8080;
+const app = express();
+const listEndpoints = require("express-list-endpoints");
```
JuliaNiki, line 10: `import listEndpoints from 'express-list-endpoints' `
you could also write like this, but your way is also correct!

server.js
```diff
@@ -1,33 +1,146 @@
...
-  res.send('Hello world')
-})
+app.get("/", (req, res) => {
+  res.send(listEndpoints(app));
+});
+// console.log(listEndpoints(app));
+//Golden Globe routes
+// const listEndpoints = require("express-list-endpoints");
+app.get("/nominations", (req, res) => {
+  const page = parseInt(req.query.page);
```
JuliaNiki, line 24: Impressed that you used parseInt function! Great job

server.js
```diff
@@ -1,33 +1,146 @@
...
+      limit: limit,
+    };
+  }
+
+  if (startIndex > 0) {
+    results.previous = {
+      page: page - 1,
+      limit: limit,
+    };
+  }
```
JuliaNiki, line 44: Nice detail that you also display the number of the previous and next pages!

server.js
```diff
@@ -1,33 +1,146 @@
...
+// Books route by id
+app.get("/books/id/:id", (req, res) => {
+  const { id } = req.params;
+  const bookById = books.find((book) => book.bookID === +id);
+
+  if (!bookById) {
+    res.status(404).json("No book found with this ID");
+  } else {
+    res.json(bookById);
+  }
```
JuliaNiki, line 113: I think this logic is also okay, but maybe it is better to first provide the case when bookById exists, else  - if not
```
if (bookById) {
   res.json(bookById)
  } else {
    res.status(404).json("No book found with this ID")
  }
```

JuliaNiki (approved):
Helena, you've done a great job! 
I really like the idea that you imported two datasets and tried to create endpoints for both of them! 
I like that you used parseInt function in this project practicing JavaScript even more.
I would only recommend to add a description/readme to this project on Github 👍🏼

PR #255, opened 2021-12-15
Drag Race Data by Anna Thunberg

annathunberg:


index.html
```diff
@@ -0,0 +1,45 @@
...
+    <link rel="shortcut icon" href="kandy.png" type="image/x-icon" />
+    <link rel="stylesheet" href="style.css" />
+    <title>drag race data</title>
+  </head>
+  <body>
+    <div
+      class="gif"
+      style="width: 100%; height: 0; padding-bottom: 56%; position: relative"
+    >
+      <iframe
```
waliem, line 23: I'm impressed that you sent an HTML file into the API to display as the front page!

index.html
```diff
@@ -0,0 +1,45 @@
...
+
+    <a
+      href="https://giphy.com/gifs/rupaulsdragrace-season-13-drag-race-rpdr-5c0kPhSU0yRTDsNBjk"
+      >via GIPHY</a
+    >
+
+    <h1>seasons of rupauls drag race</h1>
+
+    <p>please add a <span>path</span> to view different endpoints</p>
+    <p><span>/seasons</span> to view all data</p>
```
waliem, line 42: well done that you added the instructions here.

server.js
```diff
@@ -1,33 +1,77 @@
...
+const app = express();
+
+// Enables cors and json body parsing
+app.use(cors());
+app.use(express.json());
+
+// Start route on default port
+//sending a html-file as the response by using sendFile
+app.get("/", (req, res) => {
+  res.sendFile(__dirname + "/index.html");
```
waliem, line 16: I didn't know that you could do it this way, thanks for the new knowledge!

server.js
```diff
@@ -1,33 +1,77 @@
...
+});
+
+//this is a dummy
+
+app.get("/compute/complex", (req, res) => {
+  res.json({ result: 1 + 1 });
+});
+
+// error if the path does not exist in the api, * means all other paths than the ones listed above get this message
+app.get("*", (req, res) => {
```
waliem, line 63: very nice with this solution for the paths that are not valid!

server.js
```diff
@@ -1,33 +1,77 @@
...
+  if (!season) {
+    res.status(404).send("no season found with that id");
+  } else {
+    res.json(season);
+  }
+});
+
+//this is a dummy
+
+app.get("/compute/complex", (req, res) => {
```
waliem, line 58: Not sure I understood this part? I understand it's a dummy but would be interesting to know a bit more. :)

waliem (commented):
Well done Anna! You did a great job adding what I think is your own data and were able to do a lot of different stuff to your api! I like that you went the extra mile to add a "front page" by using css and html sent through as a file in the api route. 

you covered most reqs and should be proud.
Good job!

PR #270, opened 2022-05-04
Savannah Hayes Astronaut Api

savannah-hayes:


Nosslexa (commented):
Wow Savannah! Im so impressed by this project. Really nice to se how you not only made it possible to use query parameters and choose from a bunch of different routes, you even made dokumentation and a frontend for the project. 
Its also nice to se how you have used pagination, i like that you change the number of results displayed to 52 to make the frontend work as you want it to. 
Really great job this week!

PR #272, opened 2022-05-05
Express API project

rawisou:
# Project Express API
In this week's project, I built my first api using nodeJS with the Express framework. 

Viewers should be able to see the semantic endpoints and filter specific data.

**/books** is the base endpoint. Users can get all the books that exist in the provided data.

**/books/top-rated** will bring users to all books whose average rating is higher than 4

**/books/title/:title** will bring users to a specific book with a specific title. _Try "Life of Pi"_ (case insensitive) :) 

**/books/authors/:authors** will bring users to all books from a specific author. _Try "Dan Brown"_ (case insensitive) :) 

## View the API

https://express-project-deployment.herokuapp.com/


routes/books.js
```diff
@@ -0,0 +1,48 @@
...
+
+router.get("/title/:title", (req, res) => {
+    const title = req.params.title
+    let titleInput = booksData.find((str) => str.title.toLowerCase() === title.toLowerCase())
+
+    if (titleInput) {
+        res.status(200).json({
+            data: titleInput,
+            success: true
+        })
```
savannah-hayes, line 23: I noticed that when I try to find any of the Harry Potter books, I hit your else statement. I can't seem to be able to hit this endpoint for example: /books/title/Harry Potter and the Order of the Phoenix (Harry Potter  #5)

routes/books.js
```diff
@@ -0,0 +1,48 @@
...
+    } else {
+        res.status(400).json({
+            data: "Not found",
+            success: false
+        })
+    }
+})
+
+router.get("/author/:authors", (req, res) => {
+    const authors = req.params.authors
```
savannah-hayes, line 33: Nice job on creating an endpoint to get a list of books from a specific author!

server.js
```diff
@@ -1,30 +1,22 @@
...
 
-// Add middlewares to enable cors and json body parsing
 app.use(cors());
 app.use(express.json());
+app.use('/books', books)
 
-// Start defining your routes here
 app.get("/", (req, res) => {
-  res.send("Hello Technigo!");
+  res.send(listEndpoints(app));
```
savannah-hayes, line 17: Good job listing all of your endpoints in your root path.

routes/books.js
```diff
@@ -0,0 +1,48 @@
...
+        res.status(200).json({
+            data: authorsInput,
+            success: true
+        })
+    } else {
+        res.status(200).json({
+            data: "Not found",
+            success: false
+        })
+    }
```
savannah-hayes, line 45: I tested out this endpoint in postman and I never seem to hit this else statement because the if statement on line 35 always seems to be true no matter what you type, even if you type an author that doesn't exist. 

Is there a reason why you would want to send the same 200 error code and the string "Not found" in the response body?

savannah-hayes (approved):
Good job this week creating an API, you seem to meet all of the requirements as you have more than two routes, you have a /books endpoint returning an array of all books, and the /title/:title endpoint that returns a single book by its title. Nice work!

routes/books.js
```diff
@@ -0,0 +1,48 @@
...
+    const title = req.params.title
+    let titleInput = booksData.find((str) => str.title.toLowerCase() === title.toLowerCase())
+
+    if (titleInput) {
+        res.status(200).json({
+            data: titleInput,
+            success: true
+        })
+    } else {
+        res.status(400).json({
```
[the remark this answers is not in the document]
rawisou: That must have been a typo! Thank you so much Savannah. I'll correct this now.

routes/books.js
rawisou: Good question, I need to look into this. The reason why I wanted to send the 200 error code is because of the .filter. As I remember when you explained in the demo, it's a success request and the author name might be there in the future? I also wanted to return a collection of books with this author so filter seems to fit rather than find.

Thanks for pointing this out though. I will spend more time on this :)

rawisou: Fixed!! I just changed line 35 to (authorsInput.length > 0) :D 
The problem was that `authorsInput.` always return an array that's why it's always true. With that said, I checked the length of it instead hehe

savannah-hayes: Yes! Sending a 200 like how Maksy suggested is what I would do as well in this situation. Sorry if I wasn't clear, I was just curious because usually a 200 status code would actually mean that the request was successful for example but that it would just return an empty array.

I think you could just omit the else statement if you still want to send a status code of 200 and successful true as the response would be successful but just return an empty array.

savannah-hayes: Great job fixing line 35! Looks good! That makes sense..

PR #274, opened 2022-05-05
Week 17 - Jennifer WK 

jenniferwvng:
Deployed project: https://express-api-proj.herokuapp.com

### DIFF
--- a/server.js
+++ b/server.js
@@ -1,13 +1,7 @@
 import express from "express";
 import cors from "cors";
 
-// If you're using one of our datasets, uncomment the appropriate import below
-// to get started!
  import avocadoSalesData from "./data/avocado-sales.json";
-// import booksData from "./data/books.json";
-// import goldenGlobesData from "./data/golden-globes.json";
-// import netflixData from "./data/netflix-titles.json";
-// import topMusicData from "./data/top-music.json";
 
 // Defines the port the app will run on. Defaults to 8080, but can be overridden
 // when starting the server. Example command to overwrite PORT env variable value:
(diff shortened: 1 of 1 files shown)


server.js
```diff
@@ -21,7 +15,78 @@ app.use(express.json());
...
+
+  //om den har hittat angivet id som matchar ett elements id i arrayen med valda regioner, kör res.status(200)
+  if (idPerRegion) {
+    res.status(200).send({
+      data: idPerRegion,
+      success: true
+    });
+  } else {
+    res.status(200).send({
+      data: 'No region with such id found',
```
ariallahyar, line 86: Good that you make a separate response for when an ID doesn't exist. Suggestion is to revise it to status code 404 which should be used when a user tries to find one specific data point that does not exist. I would consider revising the error message as well to be the reverse of what's written because the endpoint is about the ID not existing in that specified region. 
```
  res.status(404).json({
    data: "No ID in the specified region found",
    success: false,
  });

File: server.js [hunk 2]
```diff
@@ -21,7 +15,78 @@ app.use(express.json());
...
+  } else if (dataPerRegion.length === 0) {
+    res.status(200).send({
+      data: 'Response is successful but no region with that name exists',  
+      success: true
+    });
+  } else {
+    res.status(404).send({
+      data: 'Not found',
+      success: false
+    });
```
ariallahyar, line 61: I think this `else` code will never be used. I think the comparison can be tweaked a little as well. Either the data (array) has a length greater than 0 which also means it exists, or it is an empty array. Tried testing a bit and can't get a scenario where it runs. So I think the code can be revised to just and `if-else` using the successful but empty response, removing the "Not found" In lecture I think they said we can think of Not Found being for when we're filtering, vs when we're finding like you do later in your ID.
```
  if (dataPerRegion.length !== 0) {
    res.status(200).send({
      data: dataPerRegion,
      success: true,
    });
  } else {
    res.status(200).send({
      data: "Response is successful but no region with that name exists",
      success: true,
    });
  }

ariallahyar (commented):
Nice work Jennifer! Hope you enjoyed building your first API. :)

PR #280, opened 2022-05-06
project express api Hanna

HWallberg:
https://project-express-api-hwallberg.herokuapp.com/

### DIFF
--- a/server.js
+++ b/server.js
@@ -1,17 +1,8 @@
 import express from "express";
 import cors from "cors";
 
-// If you're using one of our datasets, uncomment the appropriate import below
-// to get started!
-// import avocadoSalesData from "./data/avocado-sales.json";
-// import booksData from "./data/books.json";
-// import goldenGlobesData from "./data/golden-globes.json";
-// import netflixData from "./data/netflix-titles.json";
-// import topMusicData from "./data/top-music.json";
+import netflixData from "./data/netflix-titles.json";
 
-// Defines the port the app will run on. Defaults to 8080, but can be overridden
-// when starting the server. Example command to overwrite PORT env variable value:
-// PORT=9000 npm start
 const port = process.env.PORT || 8080;
 const app = express();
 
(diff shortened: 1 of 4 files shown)


server.js
```diff
@@ -1,29 +1,50 @@
...
+    ? res.status(200).json({ data: streamByYear })
+    : res.status(400).json({ error: "Not found" })
+})
+
+app.get("/stream/title/:title", (req, res) => {
+  const { title } = req.params;
+  const streamByTitle = netflixData.find(stream => stream.title.toLowerCase() === title.toLowerCase())
+  streamByTitle
+    ? res.status(200).json({ data: streamByTitle })
+    : res.status(400).json({ error: "Not found" })
```
Thereese, line 36: Clean setup of the ternary conditional operators with error-handling, nice!

server.js
```diff
@@ -1,29 +1,50 @@
...
+})
+
+app.get("/stream/type/:type", (req, res) => {
+  const { type } = req.params;
+  const streamByTitle = netflixData.filter(stream => stream.type.toLowerCase() === type.toLowerCase())
+  streamByTitle
+    ? res.status(200).json({ data: streamByTitle })
+    : res.status(400).json({ error: "Not found" })
```
Thereese, line 44: I couldn´t make the error-handling work on the endpoints where you used filter() method. Tried changing a bit in your code and it worked if I changed the condition (streamByTitle on row 42 in this case) to "streamByTitle.length === 0" and the switch the places of the respons. Tried on the other filtered aswell and it worked. Not sure why, but it might have to do with that the filter method returns an array so it is the length that needs to be compared. Try it out if you like!

server.js
```diff
@@ -1,29 +1,50 @@
...
 
 // Add middlewares to enable cors and json body parsing
 app.use(cors());
 app.use(express.json());
 
+
 // Start defining your routes here
 app.get("/", (req, res) => {
-  res.send("Hello Technigo!");
+  res.send("Netflix & chill!");
```
Thereese, line 16: I like the title!🍿

Thereese (approved):
The code is easy to follow and your endpoints are working great! Just a bug with some of the errors, so I have a suggestion for you in the comments. Great job!

PR #283, opened 2022-05-06
project-api-bookrating Frida 

Nosslexa:
view it here: https://project-express-api-bookrating.herokuapp.com/

### DIFF
--- a/server.js
+++ b/server.js
@@ -1,40 +1,79 @@
 import express from "express";
 import cors from "cors";
 
-// If you're using one of our datasets, uncomment the appropriate import below
-// to get started!
-// import avocadoSalesData from "./data/avocado-sales.json";
 import booksData from "./data/books.json";
-// import goldenGlobesData from "./data/golden-globes.json";
-// import netflixData from "./data/netflix-titles.json";
-// import topMusicData from "./data/top-music.json";
 
-// Defines the port the app will run on. Defaults to 8080, but can be overridden
-// when starting the server. Example command to overwrite PORT env variable value:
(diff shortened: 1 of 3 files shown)


server.js
```diff
@@ -1,27 +1,80 @@
...
 app.use(express.json());
 
-// Start defining your routes here
 app.get("/", (req, res) => {
-  res.send("Hello Technigo!");
+  res.send(listEndpoints(app));
+});
+
+//1. endpoint that returns the whole array of book ratings.
+app.get("/bookratings", (req, res) => {
```
sofiaringstedt, line 18: I saw some people using /api/bookratings here, an I thought it was a nice way to indicate the start of the api enpoints

server.js
```diff
@@ -1,27 +1,80 @@
...
+  res.status(200).json({
+    data: booksData,
+    success: true,
+  });
+});
+
+// 2. endpoint that returns one rated book by its title.
+app.get("/bookratings/title/:title", (req, res) => {
```
sofiaringstedt, line 26: I wonder if you have to have /bookratings/ for this specific endpoint? Genuine questions that I am curious about! :)

sofiaringstedt (commented):
Good job this week, your api seems to work as it should and meet the requirements! One thing I would look at is how one reaches endpoints for book titles with symbols and numbers in them ( such as Harry Potter and the Half-Blood Prince (Harry Potter  #6) ). There were quite a few items in your api that looked like this and I did not manage go get to their specific endpoint! :)

PR #284, opened 2022-05-06
Week 17 - my first API by Josefin Eliasson

Josse79:
I found this first week with backend really interesting. 
To finally get some sort of bigger perspective of how an API is built.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,13 +1,11 @@
 # Project Express API
 
-Replace this readme with your own information about your project.
-
-Start by briefly describing the assignment in a sentence or two. Keep it short and to the point.
+To build a RESTful API with node.js, containing endpoints giving arrays and unique objects. Deployment in Heroku.
 
 ## The problem
 
-Describe how you approached to problem, and what tools and techniques you used to solve it. How did you plan? What technologies did you use? If you had more time, what would be next?
+Started out to build a couple of endpoints with PAF params. When that was solved I continued to apply endpoints with query. Lastly making a definition of routes i the landing page. I built it in small steps to be able to troubleshoot every new endpoint. I used both console and Postman to check my API. If I had more time I would definitely try build a frontend with sorting structure.
 
 ## View it live
 
-Every project should be deployed somewhere. Be sure to include the link to the deployed project so that the viewer can click around and see what it's all about.
+https://technigo-express-books-api.herokuapp.com/
--- a/server.js
+++ b/server.js
@@ -23,8 +23,22 @@ app.use(express.json())
 
 // Start defining your routes here
 app.get("/", (req, res) => {
-  res.send("Hello, here is an api of random famous books")
+  const landingPage = {
+     Hello: 
+     "here is an api of random famous books",
+     Routes: [{
+       "/books": "Get whole array of books",
+       "/books/id/'idnumber'": "Get unique book by id",
+       "/books?id='idnumber'": "Get unique book by id",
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -1,13 +1,11 @@
...
 
 ## The problem
 
-Describe how you approached to problem, and what tools and techniques you used to solve it. How did you plan? What technologies did you use? If you had more time, what would be next?
+Started out to build a couple of endpoints with PAF params. When that was solved I continued to apply endpoints with query. Lastly making a definition of routes i the landing page. I built it in small steps to be able to troubleshoot every new endpoint. I used both console and Postman to check my API. If I had more time I would definitely try build a frontend with sorting structure.
 
 ## View it live
 
-Every project should be deployed somewhere. Be sure to include the link to the deployed project so that the viewer can click around and see what it's all about.
+https://technigo-express-books-api.herokuapp.com/
```
mathildakarlsson, line 11: Great, clear ReadMe!

server.js
```diff
@@ -1,28 +1,108 @@
...
+       "/books/id/'id number'": "Get unique book by id",
+       "/books?id='id number'": "Get unique book by id",
+       "/books/author/'name of author": "Get list of books from particular author",
+       "/books?author='name of author": "Get list of books from particular author",
+       "/books/title/'title of book": "Get books by title",
+       "/books?title='title of book": "Get books by title",
+     }]
+   }
+  res.send(landingPage)
+})
```
mathildakarlsson, line 29: Nice landing page with good information for the visitor. I tested the endpoints and they all seem to work :D I wasn't very familiar with the endpoints with "?" but I will definitely read a bit more about that to see if I can implement it in my project as well.

server.js
```diff
@@ -1,28 +1,108 @@
...
+
+//Routes with query
+//query example in browser http://localhost:8080/books?title=galaxy&author=douglas adams&id=377
+app.get("/books", (req, res) => {
+  const { id, author, title } = req.query
+  
+  let allBooks = booksData
+  
+  if (id) {
+    allBooks = allBooks.filter((book) => book.bookID === +id)
```
mathildakarlsson, line 39: I think you can use the "find"-method here, since it is only one book that is coming back when we use their unique id 😊

server.js
```diff
@@ -1,28 +1,108 @@
...
+    .includes(title.toLowerCase()))
+  }
+  res.status(200).json ({
+    data: allBooks,
+    success: true,
+  })
+})
+
+
+//Routes with Params
```
mathildakarlsson, line 58: I believe that you can use either Params or Query-routes and that you don't need both of them, but I'm not entirely sure of this so don't "take me on my word" 😁 But I also understand if you want to keep them both because it's two different ways of writing the endpoints and that can be good to have practiced and know for future projects!

server.js
```diff
@@ -1,28 +1,108 @@
...
+app.get("/books/id/:id", (req, res) => {
+  const { id } = req.params
+
+  const bookById = booksData.find((book) => book.bookID === +id);
```
mathildakarlsson, line 62: Yes, here you've used "find" for the id!

mathildakarlsson (commented):
Great job Josefin! Your API works perfectly and you've used different endpoints, both those that are giving the user an array and also the id-endpoint where the user gets one item back. You've definitely met the requirements and your code is very structured and looks neat! Well done!!

PR #290, opened 2022-05-06
project-express-api Amanda Tange

amandatange:


server.js
```diff
@@ -1,30 +1,99 @@
...
+      "Queries": [
+        {
+          "/books?author=string": "Get all books whose author contains the string provided",
+          "/books?title=string": "Get all books whose title contains the string provided",
+        }
+      ]
+    }
+  );
+});
+
```
Kras053, line 32: Really neat with this overview on the landing page! All the paths works wonderfully and great that with author and title you can use different strings for the search and do not need to type the entire title/authors name.  ✨📚

A small suggestion is that for consistency the path to id could be defined as "/books/bookID/:id" or similair, and for ratings "/books/ratings/:rating:", and I wrote in my explanation like this : 
"/netflixShows/release_year/'year of release'": "Get array of shows from specific release year",
but this is a really small thing : )

server.js
```diff
@@ -1,30 +1,99 @@
...
+      data: "Sorry we couldn't find book you seek",
+      success: false
+    });
+	} else {
+		res.status(200).json({
+      data: filteredBooks,
+      success: true
+    })
+	};
+});
```
Kras053, line 61: Awesome that you tried the queries also! I feel inspired to do them this week 😀 
And very nice with an informative error message. 
I think a "the" is missing here "Sorry we couldn't find the book you seek".

server.js
```diff
@@ -1,30 +1,99 @@
...
+
+app.get("/books/:id/", (req, res) => {
+  const bookByID = booksData.find(
+    (book) => book.bookID === parseInt(req.params.id));
+
```
Kras053, line 66: Looks great! I used .toString() for my year of realese const so will read up on the parseInt to learn the difference!

Kras053 (commented):
Amazing job with the project this week! Your code has a great structure and is easy to follow. It meets the blue requirements and really nice that you also used the .includes() for a better search experience! Really well done!

PR #293, opened 2022-05-06
Week 17 – Build your first API – Emma Lindell

emmajosefina:
https://emmas-books.herokuapp.com/


server.js
```diff
@@ -1,30 +1,111 @@
...
+    }
+})
+
+// Res: shows top rated to lowest rated
+// Using slice() to show top 10 out of all rated books
+app.get('/books/top-rated', (req, res) => {
+  const bookRating = books.sort(
+    (a, b) => b.average_rating - a.average_rating
+  )
+  res.json(bookRating.slice(0, 50))
```
idanaslund, line 69: This one does not seem to work. Maybe it has to do with the numbers in the slice?

server.js
```diff
@@ -1,30 +1,111 @@
...
+})
+
+// :bookID = param to trigger a param example console(req params)
+// Will look for a specific param, in this case a specific ID from books
+app.get('/books/book/:bookID', (req, res) => {
+const { bookID } = req.params
+const { book } = books.find((item) => item.bookID === +bookID)
+
+if (!book) {
```
idanaslund, line 78: This one does not seem to work, it only tells me there is no book with that ID even though I chose one I knew existed (ID nr 3). When I tried an ID I knew did not exist (ID nr 5) it tells me "internal server error".

server.js
```diff
@@ -1,30 +1,111 @@
...
+} else {
+    res.status(200).json({
+      data: bookID,
+      success: true,
+    })
+  }
+})
+
+// Random book
+app.get('/random-book', (req, res) => {
```
idanaslund, line 92: This one works and is fun, it is like getting a random tip for a book to read!

idanaslund (approved):
I really like your API and the random book was a fun extra feature! I was inspired by your different endpoints. Look in to the ones that does not work and it will be a great API to keep working with!

PR #294, opened 2022-05-07
Express API - Jessica Sandler

jessand77:
https://swedish-football-players-api.herokuapp.com/

### DIFF
--- a/server.js
+++ b/server.js
@@ -73,11 +73,13 @@ app.get('/players/position/:position', (req, res) => {
   } else {
     res.status(404).json({
       data: "Not found",
-      success: true,
+      success: false,
     })
   }
   
 })
+
+
 
 app.get('/players/random', (req, res) => {
 
@@ -90,7 +92,7 @@ app.get('/players/name/:name', ( req, res ) => {
   const { name } = req.params;
 
   const playerByName = players.find(
-    (player) => player.name === name
+    (player) => player.name.toLowerCase() === name.toLowerCase()
   );
 
   if (playerByName) {
--- a/server.js
+++ b/server.js
@@ -17,15 +17,15 @@ app.use(express.json());
 app.get("/", (req, res) => {
 
   const documentation = {
-    "Welcome": "An API about Sweden's football team",
+    "About": "This is an open API about the Sweden women's national football team",
     "Routes": [
       {
-        "/players": "List all players",
-        "/players/goals/descending": "List players sorted by number of goals in descending order",
-        "/players/games/descending": "List players sorted by number of games in descending order",
-        "/players/position/:position": "List all players in the specified position",
-        "/players/random-player": "Show info about a random player",
-        "/players/name/:name": "Show info about a specific player",
+        "/players": "Get all players",
+        "/players/goals/descending": "Get all players sorted by number of goals in descending order",
+        "/players/games/descending": "Get all players sorted by number of games in descending order",
+        "/players/position/${position}": "Get all players in a specific position (goalkeeper, defender, midfielder or forward)",
+        "/players/random": "Get random player",
+        "/players/name/${name}": "Get a specific player by name",
       }
     ]
   }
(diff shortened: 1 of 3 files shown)


server.js
```diff
@@ -21,9 +15,98 @@ app.use(express.json());
...
+  const documentation = {
+    "About": "This is an open API about the Sweden women's national football team",
+    "Routes": [
+      {
+        "/players": "Get all players",
+        "/players/goals/descending": "Get all players sorted by number of goals in descending order",
+        "/players/games/descending": "Get all players sorted by number of games in descending order",
+        "/players/position/${position}": "Get all players in a specific position (goalkeeper, defender, midfielder or forward)",
+        "/players/random": "Get random player",
+        "/players/name/${name}": "Get a specific player by name",
```
EmmaaBerg, line 28: Nice that you have your routes from less specific to more specific

server.js
```diff
@@ -21,9 +15,98 @@ app.use(express.json());
...
+      }
+    ]
+  }
+
+  res.send(documentation);
 });
 
+app.get('/players', (req, res) => {
```
EmmaaBerg, line 36: Good naming convention for the route

server.js
```diff
@@ -21,9 +15,98 @@ app.use(express.json());
...
+  res.status(200).json({
+    numberOfPlayers: players.length,
+    players: players,
```
EmmaaBerg, line 39: Easy to follow and understand what is sent back

server.js
```diff
@@ -21,9 +15,98 @@ app.use(express.json());
...
+  });
+})
+
+app.get('/players/goals/descending', ( req, res ) => {
+
+  players.sort((a, b) => b.goals - a.goals)
```
EmmaaBerg, line 45: Like that you have used `.sort()` for displaying the data

server.js
```diff
@@ -21,9 +15,98 @@ app.use(express.json());
...
+  players.sort((a, b) => b.games - a.games)
+
+  res.status(200).json(players)
+
+})
+
+// List all players in a specific position
+app.get('/players/position/:position', (req, res) => {
+
+  const { position } = req.params
```
EmmaaBerg, line 62: Nice that you have destructured the path param :D

server.js
```diff
@@ -21,9 +15,98 @@ app.use(express.json());
...
+
+  if (playersByPosition.length > 0) {
+    res.status(200).json({
+      data: playersByPosition,
+      success: true,
+    })
+  } else {
+    res.status(404).json({
+      data: "Not found",
+      success: true,
```
EmmaaBerg, line 76: Should this be `success: false`  since its status 404? Or is this the case where it's not a unique endpoint and the response is successful but just no match for the search?

server.js
```diff
@@ -21,9 +15,98 @@ app.use(express.json());
...
+    })
+  }
+  
+})
+
+app.get('/players/random', (req, res) => {
```
EmmaaBerg, line 82: Fun endpoint! :D

server.js
```diff
@@ -21,9 +15,98 @@ app.use(express.json());
...
+  const randomPlayer = players[Math.floor(Math.random()*players.length)]
+
+  res.status(200).json(randomPlayer)
+})
+
+app.get('/players/name/:name', ( req, res ) => {
+  const { name } = req.params;
+
+  const playerByName = players.find(
+    (player) => player.name === name
```
EmmaaBerg, line 93: Maybe it could be good to think about case sensitivity and add `.toLowerCase()` so it doesn't matter if the user types the name with lower or uppercase letters :)

EmmaaBerg (commented):
Good job with creating your own data set and endpoints. Liked that you used different ways, like `.sort()`, `.find()` and `.filter()` to present the data. Added a comment about case sensitivity but overall really well done this first week with backend! :D

server.js
jessand77: Yes, good idea, I forgot that. Added it now :-)

server.js
jessand77: Yes, you're right, it should be false, not true.

PR #301, opened 2022-05-08
Week 17 - Project Express API by Jenny Fiskaare

JenFi72:


server.js
```diff
@@ -1,13 +1,7 @@
-import express from "express";
+import express, { request, response } from "express";
```
CamillaHallberg, line 1: I think you can actually omit the 'request' and 'response' from your import :)

server.js
```diff
@@ -21,9 +15,117 @@ app.use(express.json());
...
+{
+      "books": "All books available", 
+      "/books/id/:id": "Search for a book by specific id",
+      "/books/title/:title": "Search for a book by specific title",
+      "/books/authors/:authors": "Search for book/books written by specific author (alone or with others)",
+      "/books/language_code/:language_code": "Search for books available in specific language",
+    }
+    ]
+  } );
 });
```
CamillaHallberg, line 29: I really enjoyed your "welcome" page! Really nice of you to specify each endpoint you have created. Considering there is no frontend on this project, I would say this is the most user friendly thing to do :) Nice and easily read from my perspective!

server.js
```diff
@@ -21,9 +15,117 @@ app.use(express.json());
...
+    data: books,
+    success: true,
+  });
+  });
+
+
+  //Endpoint 1: Searchning for a specific book by bookid
+app.get('/books/id/:id', (req, res) => {
+  const { id } = req.params
+  const bookById = books.find ((item) => item.bookID === +id)
```
CamillaHallberg, line 43: A little typo but I think you should not have a space between find method and the parenthesis starting. Other than that. Nice code! :) I tried it out and I'm able to get a book by typing an ID and if there is no valid ID I get the right response.

server.js
```diff
@@ -21,9 +15,117 @@ app.use(express.json());
...
+
+  if (!bookById) {
+    res.status(400).json({
+      ressponse: "There is no book with that ID number",
```
CamillaHallberg, line 47: Teenytiny typo here to maybe remove the extra 's' in response :)

server.js
```diff
@@ -21,9 +15,117 @@ app.use(express.json());
...
+    })
+  } else {
+    res.status(200).json({
+      response: bookById, 
+      success: true,
+    })
+  }
+})
+
+  //Endpoint 2: Searching for a specific  book title
```
CamillaHallberg, line 58: Great with a comment where your next endpoint is "being born" :)

server.js
```diff
@@ -21,9 +15,117 @@ app.use(express.json());
...
+    
+    if (!bookByTitle){
+      res.status(200).json({
+        data: "Succesful search, but that book title is misspelled or not available here.",
+        success: false
+      });
+    } else {
+      res.status(200).json({
+        data: bookByTitle, 
+        success: true,
```
CamillaHallberg, line 72: I'm not sure you should have the same status code on line 65 and line 70? I'm thinking the more "correct" way is to throw some sort of 400-status (404) on line 65?

server.js
```diff
@@ -21,9 +15,117 @@ app.use(express.json());
...
+      res.status(200).json({
+        data: filteredAuthor,
+        success: true,
+      });
+    };
+    });
+  
+
+    //Endpoint 4: Searching for all books in  a specific language  
+    app.get('/books/language_code/:language_code', (req, res) => {
```
CamillaHallberg, line 103: Oh, I like that the user can search for a book from a specific language! Although, one thing I thought of was how will the user know all the available language-codes? For example, if I type in 'English' there will be no data for English book titles. If you were to improve this by connecting with frontend an idea could be to have a dropdown menu with all available language-codes the user can choose from :) But I totally acknowledge that it was the black level requirements for this week to use frontend. 
But just a thought, from the user perspective. :)

CamillaHallberg (approved):
Great job with the project this week, Jenny! I really enjoyed your "landing" page where I get introduced to the book route and the different endpoints you created. This looks very good, nice and structured. I like your choices of endpoints to choose from and I have only made a few comments in your code in regards to the language codes, that is just something to have in mind for future projects :) 

Great job with this first week of backend, Jenny! ⭐️

PR #302, opened 2022-05-08
blue level

JensBergqvist:
I had a Terrible week this week. Slept for max 2h per night so I've just been exhausted and really didn't have the energy to try out new things. 

https://project-express-api-jens.herokuapp.com/

### DIFF
--- a/server.js
+++ b/server.js
@@ -1,13 +1,7 @@
 import express from "express";
 import cors from "cors";
 
-// If you're using one of our datasets, uncomment the appropriate import below
-// to get started!
-// import avocadoSalesData from "./data/avocado-sales.json";
-// import booksData from "./data/books.json";
-// import goldenGlobesData from "./data/golden-globes.json";
-// import netflixData from "./data/netflix-titles.json";
-// import topMusicData from "./data/top-music.json";
+import booksData from "./data/books.json";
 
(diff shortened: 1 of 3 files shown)


server.js
```diff
@@ -1,13 +1,7 @@
 import express from "express";
 import cors from "cors";
```
id4h4lling, line 3: If you would like the fancy first page you could install `npm i express-list-endpoints` and then import it here `import getEndpoints from "express-list-endpoints`

And then on line 18

server.js
```diff
@@ -21,7 +15,43 @@ app.use(express.json());
 
 // Start defining your routes here
 app.get("/", (req, res) => {
-  res.send("Hello Technigo!");
+  res.send("For the bookworms!");
```
id4h4lling, line 18: put: `res.send(getEndpoints(app))` instead.

id4h4lling (approved):
Jens ough I know it's SO hard without sleep - BUT YOU SOLVED it. Hurray to you! 🙌

If you want to add a fancy first page you could install express list endpoints. I showed you how in the comments below. And that was all I could do to improve, because I sleept some more then you (thats a first someone beat me to it) but haven't done anything more than that. Good job this week! 
/Ida

PR #304, opened 2022-05-08
Week 17 - creating an API using Express

EmmaaBerg:


TereseClaesson (approved):
Great work! Clear and structured code. Good use of the routes naming convention. You have used the routes order in the right way.Your paths have been written from less to more specific. Keep up the good work.

PR #313, opened 2022-05-08
project-express-api

NabeelMansour:


server.js
```diff
@@ -21,7 +22,47 @@ app.use(express.json());
 
 // Start defining your routes here
 app.get("/", (req, res) => {
-  res.send("Hello Technigo!");
+  res.send("Hello Book Entusiats!");
```
SteppbySteph, line 25: Nice greeting! Another thing that could be added here is some information about what routes to type - like an overview! The routes can alternatively be added in your README file together with your own thoughts.

server.js
```diff
@@ -21,7 +22,47 @@ app.use(express.json());
...
+})
+
+app.get("/books/title/:title", (req, res) => {
+  const { title } = req.params;
+
+  const bookByName = booksData.find(
+    (book) => book.title.toLowerCase() === title.toLowerCase()
+  );
+
+  if (!bookByName) {
```
SteppbySteph, line 42: Happy to see the statuses work. Nice feature!

server.js
```diff
@@ -21,7 +22,47 @@ app.use(express.json());
...
+      success: true,
+    });
+  }
+});
+
+app.get("/books/author/:author", (req, res) => {
+  const { author } = req.params;
+
+  const booksByAuthor = booksData.filter(
+    (book) => book.authors.toLowerCase() === author.toLowerCase()
```
SteppbySteph, line 59: To make the search action return some results without the need to type entire name/title we could use the include method. It would look something like this: 
`let booksByAuthor = booksData.filter(
    (book) => book.authors.toLowerCase()
    .includes(authors.toLocaleLowerCase()));`
   This could help the user type just a part of the name and it would eventually give the user more than one result back.

SteppbySteph (commented):
This week was a bit different from what we have learnt so far. You have done an amazing job with this week's project. Your code is clean and understandable which makes it easy to follow. Fantastic, Nabeel!

PR #315, opened 2022-05-09
project-express-api

mathildakarlsson:
Heroku-deployment in process..

### DIFF
--- a/server.js
+++ b/server.js
@@ -17,13 +17,13 @@ app.get("/", (req, res) => {
         "/content/id/'id-number": "GET a specific movie/tv-show by it's unique id-number",
         "/content/title/'title of movie/show": "GET movies/tv-shows by title",
         "/content/director/'director of movie/show": "GET movies/tv-shows by director",
-        "/content/type/movies": "GET whole array of movies",
-        "/content/type/shows": "GET whole array of tv-shows",
+        "/content/type/movie": "GET whole array of movies",
+        "/content/type/show": "GET whole array of tv-shows",
       }]
     }
   res.send(welcomePage)
   });
-  
+
 
 //Endpoint with all available content from API
 
--- a/server.js
+++ b/server.js
@@ -1,52 +1,146 @@
 import express from "express";
 import cors from "cors";
-import goldenGlobesData from "./data/golden-globes.json";
-// import { json } from "express/lib/response";
-
-
-// import avocadoSalesData from "./data/avocado-sales.json";
-// import booksData from "./data/books.json";
-// import netflixData from "./data/netflix-titles.json";
-// import topMusicData from "./data/top-music.json";
-
-
+import { json } from "express/lib/response";
+import netflixData from "./data/netflix-titles.json";
 
 const port = process.env.PORT || 8080;
 const app = express();
 
-
 app.use(cors());
 app.use(express.json());
 
-const data = goldenGlobesData
-
-
 app.get("/", (req, res) => {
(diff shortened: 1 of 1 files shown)


server.js
```diff
@@ -1,30 +1,131 @@
...
-// Start the server
+
+//Endpoint with type (movies or tv-shows)
+
+app.get("/content/type/:type", (req, res) => {
+  const { type } = req.params;
+
+  const contentByType = netflixData.filter(
+    (content) => content.type.toLowerCase().includes(type.toLowerCase())
+  );
```
amandatange, line 119: the /content/type/movie or show works super well in the browswer and it gives me all the movies or all the tv shows. But since the comparison is .includes() it is also possible to enter 'content/type/o' and it will give me all the movies and tv shows since show and movie both contain an o. I would maybe suggest using '===' instead of .includes() for this particular pathparam since it will give you either movies or tv shows, and then maybe handle if someone enters anything else but movie or show :)

server.js
```diff
@@ -1,30 +1,131 @@
...
-// Start defining your routes here
 app.get("/", (req, res) => {
-  res.send("Hello Technigo!");
+    const welcomePage = {
+      Hello:
+      "Hello there! Here you'll find an API with information about 1 375 Netflix movies and TV-shows!",
+      Routes: [{
+        "/content": "GET the whole array of movies and tv-shows",
+        "/content/id/'id-number": "GET a specific movie/tv-show by it's unique id-number",
+        "/content/title/'title of movie/show": "GET movies/tv-shows by title",
```
amandatange, line 18: I think this path should either be changed to content/'title of movie/show' as is written in the code on line 66, or the path in line 66 should be changed to /content/titles/'title of movie/show. I think the latter one might be good because then it's very obvious that it is a title that comes next :)

server.js
```diff
@@ -1,30 +1,131 @@
...
+});
+
+
+//Endpoint with title of movie/tv-show
+    
+app.get("/content/:title", (req, res) => {
+  const { title } = req.params;
+
+  const contentByTitle = netflixData.filter(
+    (content) => content.title.toLowerCase() === title.toLowerCase()
```
amandatange, line 70: This may be a good use for .includes() since when searching for a title you may only know a couple of words, or you might miss the last letter in a word and it would still return the titles that contain the path param.

server.js
```diff
@@ -1,30 +1,131 @@
...
+});
+
+
+//Endpoint with director name
+
+app.get("/content/director/:director", (req, res) => {
+  const { director } = req.params;
+
+  const contentByDirector = netflixData.filter(
+    (content) => content.director.toLowerCase().includes(director.toLowerCase())
```
amandatange, line 94: great use of the .inlcudes()! It makes it a lot easier for the user I think to not have to search for the exact name of the director

server.js
```diff
@@ -1,30 +1,131 @@
...
+  );
+
+  if (contentById) {
+    res.status(200).json({
+      data: contentById,
+      success: true,
+    });
+  } else {
+    res.status(404).json({
+      data: "The movie/tv-show you're looking for is not here.. Are you sure you've entered the correct id-number?",
```
amandatange, line 56: Working nicely with the error message when it doesn't find a movie with that id!

amandatange (commented):
Great job this week! You managed to get several routes and fulfill the requirements, no problem! The thing I think might be the easiest to fix, and most helful for the user, is the instruction text for the titles route :) Also very well done with the error handling as well!

PR #319, opened 2022-05-11
Express api

sk-phan:


tiiliu (commented):
Nice work Suki! You are filling the blue requirements! 🤗  I especially like the route which gives you the books with high ratings. I left some individual comments in the code, hopefully some of them could be of use for you 💖

PR #320, opened 2022-05-12
booksData by Laura

Laura-Sjolander:


server.js
```diff
@@ -1,30 +1,109 @@
...
+  
+  res.send({
+    "Books API" : "Look for books by ID, ISBN number or title",
+    "Routes":[{
+      "/booksData": "Get all books",
+      "/booksData/isbn/:isbn": "Get a book by its ISBN number",
+      "/booksData/id/:id": "Get a book by its ID number in the database",
+      "/booksData/authors/:authors": "Get a book by its author/authors",
+    }]
+  });
```
rijad90, line 27: I like the fact that you have a small "table of content" so it's a lot easier to find your routes! Nice :)

server.js
```diff
@@ -1,30 +1,109 @@
...
+app.get('/booksData', (req, res) => {
+  res.status(200).json({
+    data: booksData,
+    sucess: true,
+  });
+});
+
+//Find books by their ID number of the array by .find() method
+app.get('/booksData/id/:id', (req, res) => {
+const { id } = req.params;
```
rijad90, line 42: great job using parmas

server.js
```diff
@@ -1,30 +1,109 @@
...
+      sucess: true,
+   });
+  };
+  });
+
+//Get the book by title by filter.() method
+app.get('/booksData/title/:title', (req, res) => {
+  const { title } = req.params;
+
+  const bookByTitle = booksData.filter((booksData) => booksData.title.toLowerCase() === title.toLowerCase()
```
rijad90, line 85: Nice work to use both filter method and includes method and convert to lower letters.

rijad90 (commented):
good job with the backend! you are a natural talant Laura :)

PR #324, opened 2022-11-17
Kaja Wilbik - week 13

KajaWilbik:


server.js
```diff
@@ -21,9 +23,55 @@ app.use(express.json());
...
+  // res.send({responseMessage: "Hello Technigo!"}); // we are send when we would like to sent translate json object
+  res.json ({responseMessage: "Hello Technigo!"}) // it is sending as a json, which makes it more secure if there is some error
 });
 
+app.get("/library", (req, res) => {
+ const { title, isbn, authors } = req.query;
+   let books = booksData;
+
+  if (title) {
+    books = books.filter(singleBook => singleBook.title.toLowerCase() === title.toLowerCase()); // to be sure that API will read all possible way of writing it
```
AmandaElvkull, line 35: Like your comments that explain the code !

server.js
```diff
@@ -21,9 +23,55 @@ app.use(express.json());
...
+
+
+  res.status(200).json ({
+    success: true,
+    message: "OK",
+    body: {
+      booksData: books
+    }
+   }) // it is sending as a json, which makes it more secure if there is some error
+
```
AmandaElvkull, line 56: A suggestion for next time is to do an if else for sending back the json, so you can send back an 404 with a message if the api can't find data. :)

server.js
```diff
@@ -21,9 +23,55 @@ app.use(express.json());
...
+  const singleBook = booksData.find((item) => {
+    return(item.bookID === + req.params.id)
+  });
+  console.log(singleBook)
+  res.json ({booksData  : singleBook}) // it is sending as a json, which makes it more secure if there is some error
+});
+
+app.get("/library", (req, res) => {
+  res.json (singleBook) // it is sending as a json, which makes it more secure if there is some error
+});
```
AmandaElvkull, line 71: Is this router needed ? It looks like it returning all data as in the first route at row 30. ☺️

server.js
```diff
@@ -21,9 +23,55 @@ app.use(express.json());
 
 // Start defining your routes here
 app.get("/", (req, res) => {
-  res.send("Hello Technigo!");
+  // res.send({responseMessage: "Hello Technigo!"}); // we are send when we would like to sent translate json object
+  res.json ({responseMessage: "Hello Technigo!"}) // it is sending as a json, which makes it more secure if there is some error
```
AmandaElvkull, line 27: To make it easier for those who want to test or use your api you can add the different endpoints in this route :D

AmandaElvkull (commented):
Nice and clean code! Easy to follow and it works with the deployment! 
Good work, added some comments what you can do next time you build an api but is just small things. 

Super duper good work 👊🏼

PR #328, opened 2022-11-18
Express API

CeciliaMichele:


server.js
```diff
@@ -1,30 +1,44 @@
...
+const port = process.env.PORT || 8080;
+const app = express();
+
+// Middlewares to enable cors and json body parsing
+app.use(cors());
+app.use(express.json());
+
+// My routes
+app.get("/", (req, res) => {
+  res.json({responseMessage: "Welcome to my page where you can find information on 500 books. Please add /books/ and then a number in between 1-500, or /languages/ and for example 'fre' to see books in french." }) // Expects an json object and will be sent accordingly. More secure.
```
malmen237, line 24: Very nice Welcome-message. Made it easier to navigate the API.

server.js
```diff
@@ -1,30 +1,44 @@
...
+// Middlewares to enable cors and json body parsing
+app.use(cors());
+app.use(express.json());
+
+// My routes
+app.get("/", (req, res) => {
+  res.json({responseMessage: "Welcome to my page where you can find information on 500 books. Please add /books/ and then a number in between 1-500, or /languages/ and for example 'fre' to see books in french." }) // Expects an json object and will be sent accordingly. More secure. 
+});  
+
+app.get("/books/:bookID", (req, res) => {
```
malmen237, line 27: I noticed that some id:s isn't there and I can't load the json-file to see if they're missing in there. 
Weirdly there is no 5, 6 and 7...

malmen237 (commented):
Nice API. 😃 

Some small comments and I am missing the full list a bit, it would be nice to get the option to see all the available titles of the API when you have the ending: /books

But all in all:
Great work! 🥳

PR #332, opened 2022-11-18
Linnéas project Express API 

LinneaWa:
https://project-express-api-znvd3ytihq-lz.a.run.app


Archkrull (commented):
Great work Linnéa!

You have an endpoint with a collection as well as an endpoint that returns a single result. 
It is also restful as it gives the status as well as status info and all of your routes work as they should. You meet all the requirements!

I see that you have added and error status for one route, and it would be great if you did it for all routes as it is easier to work with!
It seems that the routes with strings are case sensitive, so of you write with lower case letters we get an empty array. This can be solved with adding an "toLowerCase" and would be great to have in future projects!

I like your comments as it is easy to read your code and what it does and you have done a great job!

Well done! 
regards Neah

PR #333, opened 2022-11-18
David Ballester

Fonnt:
https://project-express-api-kwmzxb52za-lz.a.run.app/


server.js
```diff
@@ -19,9 +12,79 @@ const app = express();
...
+    movies = movies.filter(movie => movie.type.toLowerCase() === type.toLowerCase());
+  }
+  if (director) {
+    movies = movies.filter(movie => movie.director.toLowerCase() === director.toLowerCase());
+  }
+  if (year) {
+    movies = movies.filter(movie => +movie.release_year === +year);
+  }
+
+  res.status(200).json({
```
Scarylie, line 39: If I request this and the year doesnt exist i still get repsonse status 200 "ok". https://project-express-api-kwmzxb52za-lz.a.run.app/movies?year=2021. But I think most of our codes look like that

Scarylie (approved):
Not much to comment :) I think it looks and works as it should. Good first job on backend.

PR #334, opened 2022-11-18
Week 13, Project Express API, Petra Söderström

Petrasoderstrom1612:
Stored on Google Cloud: https://project-express-api-cxkrcpeioa-lz.a.run.app/

### DIFF
--- a/data/data.json
+++ b/data/data.json
@@ -1,0 +1,3350 @@
+[
+  {
+    "year_film": 2009,
+    "year_award": 2010,
+    "ceremony": 67,
+    "category": "Best Motion Picture - Drama",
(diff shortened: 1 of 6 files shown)


ElinSegelov (commented):
Well done. I think your code is easy to follow but it might be helpful with some code comments. I also think its a good idea to remove the code you aren't using (averagePrice filtering, and the code from Daniels lectures.
It also would be nice with some explanatory comments in your api info, what is, for example the difference between /sales and /avocadosales? What kind of state are you able to query about? 
You have a lot of endpoints and you seem to meet the requirements. Well done on the first backend project!

PR #337, opened 2022-11-18
Caroline Soderstrom's Express API

SoderstromC:
https://project-express-api-yd2d4q5kyq-lz.a.run.app

### DIFF
--- a/server.js
+++ b/server.js
@@ -84,16 +84,16 @@ app.get("/members/:id", (request, response) => {
 // EP2 Books
 
 app.get("/books", (request, response) => {
-  const { authors, title } = request.query
+  const { title, authors } = request.query
   let books = booksData
-
-  if (authors) {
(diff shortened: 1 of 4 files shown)


server.js
```diff
@@ -21,7 +23,86 @@ app.use(express.json());
...
+  let books = booksData
+
+  if (title) {
+    books = books.filter(singleBookData => singleBookData.title.toLowerCase() ===
+    title.toLowerCase())
+  }
+
+  if (authors) {
+    books = books.filter(singleBookData => singleBookData.authors.toLowerCase() === authors.toLowerCase())
+  }
```
agadman, line 97: Nicely done with the if statements and filtering here of title and author. It works fine with queries.

agadman (approved):
Great job with this backend project! It meets the requirements; It has a couple of RESTful endpoints which return either an array of data or a single item. The deployed version works fine. Also, nice and clean code. :)

PR #338, opened 2022-11-18
Project Express API by Kartika

katirum:
https://project-express-api-5dpplzryzq-lz.a.run.app/


server.js
```diff
@@ -1,29 +1,132 @@
...
+      data: "Brand not found",
+      success: false 
+    })
+  } else {
+    res.status(200).json({
+      data: selectedBrand,
+      success: true
+    });
+  } 
+});
```
NinaBerggren, line 96: https://project-express-api-5dpplzryzq-lz.a.run.app/ramen/brand/nissin  only displays one item not a list.
I think it's because you used find()-method and should use the filter()-method.
(see the ramen/country/:country endpoint which uses filter() and returns a list)
from MDN Web docs mozilla : 
The find() method is an iterative method. It calls a provided callbackFn function once for each element in an array in ascending-index order, until callbackFn returns a truthy value. find() then returns that element and stops iterating through the array.

server.js
```diff
@@ -1,29 +1,132 @@
...
+
+
+// Route to ramen based on how it is served
+// example route: /ramen/style/cup
+app.get("/ramen/style/:style", (req, res) => {
```
NinaBerggren, line 101: the same as above, also only returns one item not a list

server.js
```diff
@@ -1,29 +1,132 @@
...
+});
+
+
+// Route to sort ramen from best rating to lowest
+// example route: /ramen/best
+app.get("/ramen/best", (req, res) => {
+  const best = data.sort(
+    (a, b) => b.Stars - a.Stars
+  )
+
```
NinaBerggren, line 125: Very nice and to the point code, for showing the hole list in a useful way.

NinaBerggren (commented):
Great work Tika!
Easily made the blue level requirements, even a few of the red ones, be proud.
Well written descriptions in the code and examples of how it works in the browser. Smart selected endpoints, you made it fun to look through the ramen data =)
Just one thing,  there is a smal confusion about these 2 endpoints:
"/ramen/brand/:brand": "get list of all ramen with specific brand",
"/ramen/style/:style": "get list of all ramen with specific type of serving",
if they are supposed to return only one item make a change in the "presentation of the routs/endpoints" if they are supposed to show a list make the small change in the code. 
And again, you made it very easy to go through you ramen data by having the base Route show the information of what Routs/endpoints which is available.
keep up the good work // Nina

server.js
katirum: Thank you Nina for this comment, I used filter before but somehow I changed it to find and I don't know why honestly haha but thanks for pointing it out!

PR #347, opened 2022-11-20
Express Api booksData by Saralie

Scarylie:


miadahlgren (commented):
Super nice job! The code looks great and it works well!  Great that you added some extra routes, and walked the extra mile :)

PR #351, opened 2022-11-20
project-express-api, Neah Rockstroh

Archkrull:


server.js
```diff
@@ -19,11 +12,88 @@ const app = express();
...
+    {
+      '/books': 'Get all books',
+      '/books/:bookID': 'Get specific book by ID',
+      '/isbn/:isbn': 'Get a book`s isbn',
+      '/languages/language_code': 'Get book by language',
+      '/authors/:authors': 'Get book by author',
+      '/best': 'Get books sorted by rating, highest to lowest'
+    },
+  ],
+};
```
MariaWellander, line 28: I love this! This isn't a requirement but I love that you found a solution to the problem that the user can't see what routes there is to chose from. Fantastique!

server.js
```diff
@@ -19,11 +12,88 @@ const app = express();
...
+  
+});
+
+// Gets book by ID
+app.get("/books/:bookID", (request, response) => {
+  const singleBook = books.find((book) => {
+   return book.bookID === + request.params.bookID;
+  })
+  if(!singleBook) {
+    response.status(404).send('Error: Book not found')
```
MariaWellander, line 44: Nice that you're having error messages :)

server.js
```diff
@@ -19,11 +12,88 @@ const app = express();
...
+    } else {
+    response.status(404).send("Error: Isbn not found")
+    }
+});
+
+// books sorted by average rating, highest to lowest
+app.get("/best", (req, res) => {
+  const best = books.sort(
+    (a, b) => b.average_rating - a.average_rating
+  )
```
MariaWellander, line 92: Very nice with this sort() method! Fun idea :)

MariaWellander (commented):
## Am I able to understand the code easily? Does the code have lots of duplicates? Could it be tidied up?

I'm able to understand the code easily - good comments too! Looks very neat and clean.

## Does it work or does anything look broken?

I've tried every route and they work perfectly - good job!

## Does it follow the general and blue requirements?

Yes! It even meets up to the requirements for the red level - cool!

## To sum it up

Super good work, Neah! I think you should be very proud of yourself. Your first API - wow! I really like it that you let the user see the routes in the base path. It was also fun that you had that many routes - not just 2 ones (the minimum requirement). High five!

PR #354, opened 2022-11-20
Sarahs-GoldenGlobe-Express-API

Kneedler:


server.js
```diff
@@ -1,27 +1,66 @@
...
+/*  app.get('/globes/:release', (req, res) => {
+  const release = +req.params.year_film
+  const yearFilm = goldenGlobesData.filter((globe) => globe.year_film)
+
+  if(yearFilm) {
+    res.json({ yearFilm: yearFilm }) 
+  } else {
+    res.status(404).json(`Sorry, no release year: ${release}`)
+  }
+})  */
```
Fonnt, line 43: I encountered the same issue. I solved it by giving a new endpoint. Something like this:
`app.get('/globes/release/:release`

server.js
```diff
@@ -1,27 +1,66 @@
...
+  if (ceremonyNumber) {
+      res.status(200).json({
+      success: true,
+      message: "OK",
+      response: {
+      goldenGlobesData: ceremonyNumber
+    }
+  });
+} else {
+  res.status(404).json(`Ceremony number ${ceremonyNumber} not found`); //This part does not work
```
Fonnt, line 28: This was a really nice thought! Interesting approach to try to implement what the user typed... 
I think that you are missing that the json should be an object. You can try with this: 

```
res.status(404).json({
      success: false,
      message: "Ceremony not found",
      response: {}
    });
```

And then in the message part, try with `${ceremony}` ? or `${:ceremony}` or `${params.ceremony}` and see if it works.

server.js
```diff
@@ -1,27 +1,66 @@
...
+  if (category) {
+      globes = globes.filter(singleGlobe => singleGlobe.category.toString().toLowerCase().includes(category.toLowerCase()));
+  }
+ res.status(200).json({
+   success: true,
+   message: "OK",
+   response: {
+   goldenGlobesData: globes
+}
+});
```
Fonnt, line 63: Nice choose with this queries. They come in handy

server.js
```diff
@@ -1,27 +1,66 @@
...
 
-// Add middlewares to enable cors and json body parsing
 app.use(cors());
-app.use(express.json());
+app.use(express.json()); 
 
-// Start defining your routes here
 app.get("/", (req, res) => {
-  res.send("Hello Technigo!");
+  res.send({responseMessage: "Golden Globe stats, library at /globes"});
```
Fonnt, line 12: I used this first path to give some more information of the different endpoints. I think it would be nice to implement it. You could try with something similar to this: 

```
app.get("/", (req, res) => {
  res.status(200).json({
    resMessage: "Netflix movies",
    endPointsArray: "/movies - Returns an array of all movies",
    endPointsArrayQuery: "/movies ? type, director, year",
    endPointsSingleItem: "/movies/id/:id - /movies/title/:title"
  });
});

```

Fonnt (commented):
Hej Sarah! Good job this week! you fulfill all the requirements and it looks nice! 
I may have found why your message aren't displaying. Take a look att that comment. It could be because status.json should be an object.
I think that you had a really nice idea trying to give what the user typed as an info in the status!  
Good job and keep doing! 
/Font

server.js
Kneedler: Hej David, thanks for taking your time to review and helping me out with your great tips! Much appreciated 🫶

PR #358, opened 2022-11-20
Thérèse Project express API

Tanmark:


server.js
```diff
@@ -1,27 +1,64 @@
...
-  res.send("Hello Technigo!");
+  res.send("Hello, let's check out the world of BOOKS!")
+});
+
+// Route Two - All books with query params (title and average_rating)
+app.get("/books", (req, res) => {
+  const { authors, average_rating } = req.query
+  let books = booksData
+  if (average_rating) {
+    books = booksData.filter(rating => rating.average_rating.toLowerCase() === average_rating.toLowerCase());
```
jessikalind, line 27: I think you might need to add something to the filter to make the filtering on average_rating necessary. You could either sort from highest rating to lowest for example, or perhaps just show rating over 4 ( for example `books = books.filter((rating) => rating.average_rating > 4`)

server.js
```diff
@@ -1,27 +1,64 @@
...
+  }
+  if (authors) {
+    books = booksData.filter(name => { return name.authors.toLowerCase() === authors.toLowerCase()});
+  }
```
jessikalind, line 31: Like we discussed on the Monday morning stand-up, the return {} is perhaps not needed here since everything fits into one line.

server.js
```diff
@@ -1,27 +1,64 @@
...
+
```
jessikalind, line 32: An alternative is to use `include.` instead of `===`, which would make it possible to write out only parts of the author's name and still get the needed results from the API. 

For example:   
```
if (authors) {
    books = booksData.filter((name) =>
    name.authors.toLowerCase()
    .includes(authors.toLowerCase()))
```

server.js
```diff
@@ -1,27 +1,64 @@
...
+        books: singleBook
+      }
+    });
+  } else {
+    res.status(404).json({
+      success: false,
+      message: "Not Found",
+      body: {}
+    });
+  }
```
jessikalind, line 60: Great with the helpful error message! ☺️

server.js
```diff
@@ -1,27 +1,64 @@
...
+    books = booksData.filter(name => { return name.authors.toLowerCase() === authors.toLowerCase()});
+  }
+
+  res.status(200).json({
+    success: true,
+    message: "OK",
+    body: {
+      booksData: books
+    }
+  });
```
jessikalind, line 39: When I received my code review this week, it was pointed out that it could be helpful to add another error-msg when - for example - searching for an author that does not exist. So I'll pass this comment on to here too, since the user now only get an empty array + success:true when searching for a slightly misspelled "JK Rowlings", which can be a bit confusing.

jessikalind (approved):
Great job on this week's project! The API works well, and the different endpoints are helpful. It is only the "average_rating"-endpoint I have some questions/suggestions regarding (see more in-depth comment below). 🥳

PR #360, opened 2022-11-20
Project Express API by Fanny Cardell

fannycardell:


server.js
```diff
@@ -19,11 +20,78 @@ const app = express();
 app.use(cors());
 app.use(express.json());
 
-// Start defining your routes here
+// Start defining your routes here, the thing that you want to get from the database. To get to the route
+// in the browser you write "/" and then the name of the route 
 app.get("/", (req, res) => {
-  res.send("Hello Technigo!");
+  res.json({responsMessage: "Let us see if this message will show and if it is possible to search for book titles and authors"});
```
marwebdesign, line 26: It would be nice if the message could give some hints on which endpoints there is, so it would be easier to try them in the browser (:

server.js
```diff
@@ -19,11 +20,78 @@ const app = express();
...
+
+ /* if(authors) {
+		books.filter((item) => 
+			item.authors.toLocaleLowerCase()
+			.includes(authors.toLocaleLowerCase()))
+	}
+	if(title) {
+		books.filter((item) => 
+			item.title.toLocaleLowerCase()
+			.includes(title.toLocaleLowerCase()))
```
marwebdesign, line 59: A little cleanup here (:

marwebdesign (commented):
Hard to know what to say about these backend sprints (: I tried the endpoints in Postman and they worked great, so I guess all done and clear here (: Some cleanup maybe, in code and data you are not using. Other than that, great job Fanny!!

PR #361, opened 2022-11-20
 project express api by Linnéa Ajger

LinneaAjger:
https://project-express-api-dtqfylroda-lz.a.run.app/

### DIFF
--- a/server.js
+++ b/server.js
@@ -33,7 +33,6 @@ app.get('/books', (req, res) => {
     filteredBooks = filteredBooks.filter((item) => 
     item.authors.toLocaleLowerCase()
     .includes(author.toLocaleLowerCase()))
-    console.log(filteredBooks)
   }
 
 	if(title) {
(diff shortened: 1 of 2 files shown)


server.js
```diff
@@ -21,9 +14,105 @@ app.use(express.json());
...
+  } else {
+      res.status(200).json({
+      success: true,
+      message: "OK",
+      body: {
+        booksData: filteredBooks
+      }
+    });
+  }
+
```
fannycardell, line 59: Nice that you got this filter function to work instead of the other "if" solution that was suggested!

server.js
```diff
@@ -21,9 +14,105 @@ app.use(express.json());
...
+        booksData: singleBook
+      }
+    })
+  }
+});
+
+function paginatedResults (model) {
+  return (req, res, next) => {
+    const page = +req.query.page 
+    const limit = +req.query.limit
```
fannycardell, line 89: Great that you added this function, it's a nice feature to split up the data/info on more pages than one for the user (because that is what it's for right?) 
I'm not sure if it is possible to see here in backend?

fannycardell (commented):
Hey Linnea!
Well this weeks project was a bit different from the others, so all I can say is GREAT job with it! I tried to fill in different authors in the browser and the filter worked fine and gave back the books only from that author=)
Nice!

PR #367, opened 2022-11-20
JohannaRosenholm - express-API

RosenholmJohanna:


server.js
```diff
@@ -19,12 +13,69 @@ const app = express();
...
+// Start Route
 app.get("/", (req, res) => {
-  res.send("Hello Technigo!");
+  res.send(" Go to /endpoints to see available endpoints ");
+});
+
+//shows the available endpoints
+app.get("/endpoints", (req, res) => {
+  res.send(listEndpoints(app))
+});
```
SofieRydmark, line 27: nice with this

server.js
```diff
@@ -19,12 +13,69 @@ const app = express();
...
+
+  if (author) {
+    filteredBooks = filteredBooks
+    .filter((item) => item.authors.toLocaleLowerCase()
+    .includes(author.toLocaleLowerCase()))
+  };
+  if (title) {
+    filteredBooks = filteredBooks
+    .filter((item) => item.title.toLocaleLowerCase()
+    .includes(title.toLocaleLowerCase()))
```
SofieRydmark, line 42: were you going to use these as params or queries? :) they are not listed in the endpoints

SofieRydmark (approved):
Good job! Unsure if you wanted to add the queries for author and title under /books tho. It would be a good idea to add insctructions for that in the /endpoints :)

PR #369, opened 2022-11-20
Express-API w13 -NinaBerggren

NinaBerggren:


server.js
```diff
@@ -1,30 +1,203 @@
...
+});
+
+// Route 4 All TED Talk Events list
+// ex /events
+
+app.get("/events", (req, res) => {
+  const allEvents = tedTalkData.map((talkEvent) => talkEvent.event);
+  // New set to remove all dublicates
+  const allEventsSet = Array.from(new Set(allEvents));
+
```
joeohm, line 86: Very impressive, we haven't learned about `new Set()` in this course but this is a very good use case :)

joeohm (commented):
Good work on this week's API Nina! 🙌

I like your "roadmap" documentation for the available endpoints with the explanation of expected input.

- [x]  Am I able to understand the code easily?
- [x]  No duplicates
- [x]  Works as intended, no bugs
- [x]  Follows the general and blue requirements?

// Joel

PR #384, opened 2023-05-05
Api express

IrupePozoGraviz:


server.js
```diff
@@ -1,27 +1,117 @@
...
 
 // Defines the port the app will run on. Defaults to 8080, but can be overridden
 // when starting the server. Example command to overwrite PORT env variable value:
 // PORT=9000 npm start
+
+
 const port = process.env.PORT || 8080;
 const app = express();
-
+const listEndpoints = require('express-list-endpoints');
```
jonsjak, line 15: Nice! Is this returning all the endpoints on your starterpage ('/')?

server.js
```diff
@@ -1,27 +1,117 @@
...
+      }
+    });
+  } else {
+    res.status(404).json({
+      success: false,
+      message: "title not found",
+      body: {}
+    });
+  }
+});
```
jonsjak, line 49: This works as it should! If you add to your condition to check if showID.length is 8 characters long, then you can send different error messages for wrongful id (eg. too many/few numbers) code 400, and an error for correct id-input but no matching id (404). Just a small addition, not necessarily needed.

`if (id.length !== 8) {
    res.status(400).json({
      success: false,
      message: "Invalid ID length. The ID should be 8 characters long.",
      body: {}
    });`

server.js
```diff
@@ -1,27 +1,117 @@
...
```
jonsjak, line 70: the endpoint title/:title is not working because it gets mixed up with title/:id. Because id is read first in the code the title is treated as an id and because it's not a number it runs the 404 error for the GET '/title/:id'. To solve this you can change the endpoints to be unique like, title/:title and just id/:id or use queries like title?title=Batman&id=12345678

jonsjak (commented):
The code is well structured with relevant endpoints for different filters. The error codes helps the user to see if request was successful or not. There is a problem in the get request for title/:title endpoint which I have commented in more detail at the specific part of the code. Otherwise it works as it should and a nice touch adding the endpoints at the starter page for easy read and manuevering. Good job!

PR #386, opened 2023-05-05
giorgio Api project

GiorgioCugusi:
https://project-express-api-wqewfjb3ha-lz.a.run.app

### DIFF
--- a/server.js
+++ b/server.js
@@ -1,6 +1,6 @@
 import express from "express";
 import cors from "cors";
-import technigoMembers from "./data/technigo-members.json";
+import books from "./data/books.json"
 
 // If you're using one of our datasets, uncomment the appropriate import below
 // to get started!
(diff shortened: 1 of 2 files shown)


server.js
```diff
@@ -1,5 +1,6 @@
 import express from "express";
```
malvinamaria, line 1: It's recommended to use single quotes (') instead of double quotes (") for import statements.

server.js
```diff
@@ -21,9 +22,78 @@ app.use(express.json());
 
 // Start defining your routes here
 app.get("/", (req, res) => {
-  res.send("Hello Technigo!");
+  res.send({responseMessage: "Hi there! Type in /books to browse all books, or try /random-book to get a random book recommendation!"});
```
malvinamaria, line 25: You could add here something like this:
app.get('/', (req, res) => {
  res.json({
    responseMessage:
      'Hi there! Type in /books to browse all books, or try /random-book to get a random book recommendation!',
    data: listEndpoints(app),
  });
});
and import import listEndpoints from 'express-list-endpoints'; >> then you. will have your all endpoints listed on the "/" page

malvinamaria (commented):
Your code looks great overall!

PR #388, opened 2023-05-05
EvaSjoholm/Project-Express-API

EvaSjoholm:


noraterbocs:
I like it that you put a logic in it to search for a certain word in the titles, although you shouldn't show the total data again.
Instead of this:
```
res.status(200).json({
      data: byTitle,
       success: true,
      body: {
        booksData
       }
     });
```
     You could write this:
```
     res.status(200).json({
       success: true,
      body: {
        byTitle
       }
     });
```
     but then I guess you need to change your frontend's logic, the way you access the response json after fetching. 
     So this is just for the future to know. :) 
   I can also see that you implemented couple of stretch goals, as well!  Good job, Eva!

PR #390, opened 2023-05-05
Express API Emilia

emiliasaberski:


server.js
```diff
@@ -14,16 +16,124 @@ import cors from "cors";
...
+//       data: movies
+//     }
+//   })
+// } else {
+//   res.status(500).json({
+//     success: false,
+//     message: "Something went wrong",
+//     body: {}
+//   })
+// }
```
fannystenberg, line 42: maybe remove this code since it's commented out?

server.js
```diff
@@ -14,16 +16,124 @@ import cors from "cors";
...
 });
 
+app.get('/movies', (req, res) => {
+  res.json(data)
```
fannystenberg, line 46: is there any specific reason to why you don't have a status and error message for this response?

server.js
```diff
@@ -14,16 +16,124 @@ import cors from "cors";
...
+      }
+    })
+  } else {
+    res.status(404).json({
+      success: false,
+      message: "Movie not found",
+      body: {}
+    })
+  }
+})
```
fannystenberg, line 135: clean code and easy to follow / read!

fannystenberg (commented):
Great work! You've done more than blue level, all endpoints work and nice that you used both find and filter and also both path and query for your routes!

PR #393, opened 2023-05-06
Michelle project API Express

smirrebinx:
https://project-express-api-kpnlmcrmoq-lz.a.run.app


edv-rd (commented):
Clean and efficient code. Easy to follow along and understand. Looks like a stable and well-functioning back end.

For usability's sake maybe add more responses, like a 404 when a title isn't found or a 500 when there's another error. I see you have already done a few which I think is good :)

Nice job!

PR #394, opened 2023-05-06
project express api

theresBL:


server.js
```diff
@@ -1,30 +1,121 @@
...
+})
+
+// Write:
+// http://localhost:8080/language/eng to find the books written on that specific language
+app.get('/language/:language', (req, res) => {
+  const language = req.params.language
+  const languageCode = booksData.filter((item) => item.language_code === language)
+  // res.json(languageCode)
+  if (languageCode.length) {
+    res.status(200).json({
```
MalinSkill, line 60: Nice work with the status code and the messages sent.

server.js
```diff
@@ -1,30 +1,121 @@
...
+      success: false,
+      message: "Sorry, we cant't find what you are looking for",
+      body: {}
+    })
+  }
+})
+
+app.get('/rating/:rating', (req, res) => {
+  const rating = req.params.rating
+  const minimumRating = booksData.filter((item) => item.average_rating >= +rating) //will show the books with the minimum rating of what you type in the endpoint
```
MalinSkill, line 78: Great comparison here, looks neat when using the +rating to convert it to Number

server.js
```diff
@@ -1,30 +1,121 @@
...
+  } else {
+    res.status(404).json({
+      success: false,
+      message: "Sorry, we cant't find what you are looking for",
+      body: {}
+    })
+  }
+})
+
+app.get('/pages/:pages', (req, res) => {
```
MalinSkill, line 97: I liked the different endpoints you have chosen and all of them works great as far as I can tell

server.js
```diff
@@ -1,30 +1,121 @@
...
-app.use(cors());
-app.use(express.json());
+app.use(cors())
+app.use(express.json())
 
 // Start defining your routes here
-app.get("/", (req, res) => {
-  res.send("Hello Technigo!");
-});
+app.get('/', (req, res) => {
```
MalinSkill, line 18: Daniel showed us a tip which makes it easy to show all the different endpoints in the '/'. I found it nice to use.
`const listEndpoints = require('express-list-endpoints')`
`app.get('/', (req, res) => {
  res.json(listEndpoints(app))
});`
Maybe something to try for next project

MalinSkill (commented):
Nice work! Very good endpoints and queries. Great structure and responses to the user of this API. I really liked that you used template literals to connect the message to the actual "search". I copied that from you in my own project, so thank you! Awesome work this week!

PR #400, opened 2023-05-06
Jennifers API project

jenster5:
API from Cloud Platform 
https://project-express-api-7co7srd3ia-lz.a.run.app

Frontend Deployed Page
https://jennifers-week-13-api.netlify.app/


server.js
```diff
@@ -1,30 +1,86 @@
...
+  res.status(200).json({
+    success: true,
+    message: "OK",
+    body:{
+      title: bookTitle
+    }
+  });
+}
+});
+
```
NinaWald, line 80: Good work on the project, you have met the requirements plus a nice frontend that uses the API.
/book-titles - returns a list of all book titles
/book-authors - returns a list of all book authors
/book-wellrated - returns a list of book titles with a rating of 4 or higher
There is also an endpoint that retrieves a single book title based on a user-provided search term:
/book-titles/:title - returns a single book title that matches the provided search term
So in total, there are four endpoints in the code, which is even better than the required minimum of two endpoints!

NinaWald (commented):
Well done on the project

PR #405, opened 2023-05-07
Project Express API

SofiaGerdmar:


camcron (approved):
Great work on the API Sofia! All the endpoints work when trying it out in the deployed version. Good working using both queries and params, as well as using both filter() and find(). I have nothing to add as everything looks great according to the requirements for the project :)

Keep up the good work!

PR #411, opened 2023-05-07
Project w 13 Express API

NinaWald:


khleoth (commented):
Well done, you have the netflix data up and running, as far as I can see there is still some functionality missing on the other endpoints, how does it look on your end when checking the deployed server from your google cloud page? I might be the one missing something when checking it. Backend is alot more depending on javascript knowledge, it's quite the challenge, keep up the good work!

PR #413, opened 2023-05-07
Project Express API Books

dannebrob:
live link: https://project-express-api-5igwqikbcq-uc.a.run.app/

### DIFF
--- a/server.js
+++ b/server.js
@@ -1,5 +1,6 @@
-import express from "express";
-import cors from "cors";
+import express from 'express';
+import cors from 'cors';
+import Data from './data/books.json';
 
 // If you're using one of our datasets, uncomment the appropriate import below
 // to get started!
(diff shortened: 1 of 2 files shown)


server.js
```diff
@@ -20,8 +21,61 @@ app.use(cors());
...
+  const lang = req.params.lang;
+  const booksInLanguage = Data.filter((item) => item.language_code === lang);
+
+  booksInLanguage.length !== 0
+    ? res.json(booksInLanguage)
+    : res.status(404).send('No books in that language found, please try again'); //if no books are found, send error message
+});
+
+// Endpoint to get all books with a specific rating and in a specific language
+app.get('/books/language/:lang/rating/:rating', (req, res) => {
```
Cheroptera, line 47: Is there a way to be able to round the numbers up and use no decimals? It makes it pretty hard for the user/developer to come up with a search for "4.56" with no info about those ratings beforehand.

server.js
```diff
@@ -20,8 +21,61 @@ app.use(cors());
...
 // Start defining your routes here
-app.get("/", (req, res) => {
-  res.send("Hello Technigo!");
+app.get('/', (req, res) => {
+  res.send('Welcome to my book API!');
+});
+
+//endpoint to get one random book
+app.get('/books/random', (req, res) => {
+  const randomBook = Data[Math.floor(Math.random() * Data.length)];
```
Cheroptera, line 30: I love this one!

server.js
```diff
@@ -20,8 +21,61 @@ app.use(cors());
 app.use(express.json());
 
 // Start defining your routes here
-app.get("/", (req, res) => {
-  res.send("Hello Technigo!");
+app.get('/', (req, res) => {
+  res.send('Welcome to my book API!');
```
Cheroptera, line 25: I love being welcomed to a page, but I also have to give you a tip about running `npm i express-list-endpoints` and then using that on the first page. Then you get a list of all possible endpoints in the code and makes it possible to just click on them to see the different lists 😄

server.js
```diff
@@ -20,8 +21,61 @@ app.use(cors());
...
+  const rating = req.params.rating;
+
+  const booksInRating = Data.filter((item) => item.average_rating === +rating);
+  booksInRating.length !== 0
+    ? res.json(booksInRating)
+    : res.status(404).send('No books with that rating found, please try again'); //if no books are found, send error message
+});
+
+// Endpoint to get top 10
+app.get('/books/top10', (req, res) => {
```
Cheroptera, line 71: Nice one! Love that I can get a top-ten list 🥳

Cheroptera (commented):
Well done Daniel! 👏 Your endpoints work well and I can see how they're all useful. Is there a reason to why you don't have one route showing all books? 🤔

PR #415, opened 2023-05-07
Project Express API

AnnikaSonnek:


OscarSindemark (approved):
Hey Annika! You went above and beyond with this one, I tried to imitate it in this weeks project but fell through entirely haha! your endpoints are restful, you have more than 2 endpoints to link to, and the order makes sense! Your // messages are always very detailed and make it easy to follow!

PR #427, opened 2023-05-09
aa

AnnaElvine:


FionaKlacar (commented):
Great work, Anna! Cool range of routes you have here - really useful ones for finding the film you want to see :) have you thought about adding some return status messages if you can't find something you're looking for? This would help out the frontend! Overall, good job!

PR #438, opened 2023-11-15
express-api-maria

mariateresepettersson:


backend/server.js
```diff
@@ -0,0 +1,99 @@
+import express from "express";
+import cors from "cors";
+import listEndpoints from "express-list-endpoints";
+
+import nobelData from "./data/nobel-data.json";
+
+console.log(nobelData);
```
Minadarabi, line 7: you can remove the log i think

backend/server.js
```diff
@@ -0,0 +1,99 @@
...
+app.get("/", (req, res) => {
+  const endpoints = listEndpoints(app);
+  res.json({ endpoints });
+});
+
+// Show all data (all winners of all time)
+// Filter on category and year (i.e. /winners?category=Peace&year=2000)
+app.get("/winners", (req, res) => {
+  let filteredWinners = [...nobelData]; //Makes a copy of the array
+  const { category, year } = req.query;
```
Minadarabi, line 29: good job in using query params

Minadarabi (commented):
really great job for the first backend project, you could make good use of filters and find specially on combination with query and path params

PR #441, opened 2023-11-15
Express API

AmandaRam:


index.js
```diff
@@ -0,0 +1,63 @@
...
+
+  if (type) {
+    shows = shows.filter((s) => s.type === type);
+  }
+
+  if (rating) {
+    shows = shows.filter((s) => s.rating === rating);
+  }
+
+  res.json(shows);
```
Sheryll-Ann, line 44: clean coding into one line two different filters

index.js
```diff
@@ -0,0 +1,63 @@
...
+});
+
+app.get("/shows/:showId", (req, res) => {
+  const showId = parseInt(req.params.showId);
```
Sheryll-Ann, line 48: well thought specifying the integer format of the show id

package.json
```diff
@@ -3,8 +3,8 @@
   "version": "1.0.0",
   "description": "Starter project to get up and running with express quickly",
   "scripts": {
-    "start": "babel-node server.js",
-    "dev": "nodemon server.js --exec babel-node"
+    "start": "babel-node index.js",
+    "dev": "nodemon index.js --exec babel-node"
```
Sheryll-Ann, line 7: what is the difference between server.js and index.js if I may?

Sheryll-Ann (commented):
Clean and optimised code writing as mentioned.  
I do have second thoughts about giving public access to the deployed backend, expecially with this week's (14) project containing post requests, and I'm relieved to see you have already given it a thought for week13's project.
Maybe you would like to share what Vercel is about? :)

AmandaRam:
Hi! I’m not exactly sure what you mean, can you elaborate? 🤗26 nov. 2023 kl. 20:24 skrev LeeAnn ***@***.***>:﻿
@Sheryll-Ann commented on this pull request.



In index.js:
> +app.use((req, res, next) => {
+  if (req.query.key === "secret-key") {
+    next()
+  } else {
+    res.status(401).json({ message: "Invalid key, add 'secret-key' as a query parameter at the end of the url like this: ?key=secret-key" });
+  }
+});

I remember seeing your message during the code session about the next param we haven't covered yet, but I do agree that the deployed version of our server.js should be protected instead of being shared online.  This week's (week 14) project includes post requests with all the risks of having abusive content implemented into our database from outsiders and consequently being sanctioned by either Github or Atlas.  Please do share your pov on that.

PR #442, opened 2023-11-16
express-api

EmmyLJ:


SVMaxica (commented):
I feel like your code looks very similar to mine so i dont have any complains here, it looks nice and tidy, but I would suggest that you provide the link to your deploy also in the pullrequest. Good job!😃

PR #486, opened 2024-05-02
Express API  Books - Linda

linda-f:
## Netlify link
https://linda-f-project-express-api.onrender.com

## Collaborators
[linda-f]


AntonellaMorittu (approved):
Hi Linda, congratulations for completing your first API! 👏 
Your code indeed meets the specified requirements, nice! 😎 Just one thing needs to be adjusted:
- Documentation Route: Your root endpoint / attempts to return documentation using `endpoints`, but `endpoints` is not defined at that point in your code. You need to move the declaration of endpoints on the top of your file before using it in the root endpoint:

```javascript
// Getting all endpoint with express List Endpoints
const endpoints = expressListEndpoints(app);

// Route for documentation
app.get("/", (req, res) => {
  res.json(endpoints);
});

``` 

Great job and keep it up 🚀

linda-f:
Hi @AntonellaMorittu
Thank you for your review.
I tried what you suggest as well in the beginning of the project, but then I get a blank page for the route to the documentation. I guess because the endpoints are not defined at this point in the code? So either I have to put it inside the route to the documentation or if I want to have it outside, it only works when I have it at the end like right now. What would be the better approach?

AntonellaMorittu:
Have you tried defining them inside the route?
```
app.get("/", (req, res) => {
  const endpoints = expressListEndpoints(app);
  res.json(endpoints);
});
```
this should work :)

linda-f:
Yes I did. 
This is what I meant with inside the route to the documentation. This works. So I guess this would be the better approach ;) Thx!

server.js
```diff
@@ -21,9 +16,56 @@ app.use(express.json());
...
+  } else {
+    res.status(404).send("Sorry, there is no book with that Id.");
+  }
+});
+
+app.get("/averageratings/:ratingNum", (req, res) => {
+  const { ratingNum } = req.params;
+
+  const rating = booksData.filter(
+    (rating) => Math.round(rating.average_rating) === +ratingNum
```
Martin-Joensson, line 56: Nice with the average rating. It would be cool if you could filter a range of ratings. like 3-5 or something. Just an idea =)
/Martin

Martin-Joensson (commented):
Nice work Linda! It looks like you meet all the requirements. 
Go team moon! 
/Martin

PR #491, opened 2024-05-03
Sofie Ferrari Strahl - Project Express API

SofieFerrari:
https://project-express-api-1-5qze.onrender.com/


server.js
```diff
@@ -1,30 +1,90 @@
...
+	res.json(endpoints)
+})
+
+//route that shows all data from jsonfile
+app.get('/songs', (req, res) => {
+	res.json(topMusicData)
+})
+
+//route to find and show song by id
+app.get('/songs/:songId', (req, res) => {
```
linda-f, line 28: Nice to make the songId param connected to the songs endpoint. Makes perfect sense for me when thinking about how the user will probably use the frontend. 🥳

server.js
```diff
@@ -1,30 +1,90 @@
...
+	} else {
+		res.status(404).send('song not found')
+	}
+})
+
+// route to map all the artists names
+app.get('/artists', (req, res) => {
+	const artistNames = topMusicData.map((song) => song.artistName)
+
+	if (artistNames.length > 0) {
```
linda-f, line 44: Cool idea with only making the artists the result of an endpoint! 👍

server.js
```diff
@@ -1,30 +1,90 @@
...
+
+	if (popular.length > 0) {
+		res.json(popular)
+	} else {
+		res.status(404).send('could not sort by popularity')
+	}
+})
+
+//route to sort by popularity and show the top ten songs by slice
+app.get('/favorites/:top10', (req, res) => {
```
linda-f, line 63: I am not entirely certain the URL should be written like this?
Because I thought the ":" signals you want to use a param and if I saw correctly when testing your endpoint there is only "top10" and no other params a user can add. So I guess the URL should be: /favorites/top10 ? 🙈 
However your code seems to work, so I could be wrong. 🤷‍♀️

server.js
```diff
@@ -1,30 +1,90 @@
...
+	if (top10Songs.length > 0) {
+		res.json(top10Songs)
+	} else {
+		res.status(404).send('could not show top 10 songs')
+	}
+})
+
+//route to filter bpm 128 = sweet spot
+app.get('/bpms', (req, res) => {
+	const sweetSpot = topMusicData.filter(
```
linda-f, line 77: Love the sweet spot idea! 🥇

linda-f (commented):
It looks like you had a lot of fun creating all your endpoints with express.js. And you also thought of some fun ones! Great job Sofie 💯

AntonellaMorittu (approved):
Hello Sofie, well done completing your first API! 👏 
Your code indeed meets the specified requirements, way to go! 🔥 One little improvement for the top 10 endoint. If the endpoint` /favorites/:top10 `always returns the top 10 songs based on popularity and the query doesn't change dynamically, it would be more appropriate to name the endpoint `/favorites/top10`.

Using a dynamic parameter in the endpoint path (:top10) suggests that the endpoint can return a variable number of top songs based on the value of the parameter. However, in this case, the endpoint consistently returns the top 10 songs, so naming it `/favorites/top10` would be clearer and more intuitive.

Great job 💪

PR #503, opened 2024-05-04
project-express-api

Paulajungaker:
## Netlify link
https://project-express-api-k96w.onrender.com/

### DIFF
--- a/server.js
+++ b/server.js
@@ -20,8 +20,15 @@ app.use(cors());
 app.use(express.json());
 
 // Start defining your routes here
+// http://localhost:8080/
 app.get("/", (req, res) => {
   res.send("Hello Technigo!");
+});
+
+// Get all topMusicData
+// http://localhost:8080/songs
+app.get("/songs", (req, res) => {
+  res.json(topMusicData);
 });
 
 // Start the server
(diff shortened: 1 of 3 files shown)


server.js
```diff
@@ -20,8 +14,43 @@ app.use(cors());
 app.use(express.json());
 
 // Start defining your routes here
+// http://localhost:8080/
 app.get("/", (req, res) => {
-  res.send("Hello Technigo!");
+  res.send("Top 50 popular songs on spotify");
+});
```
AntonellaMorittu, line 20: in here you could return a documented list of your endpoints using [Express List Endpoints](https://www.npmjs.com/package/express-list-endpoints), which is also a basic requirement 👍

AntonellaMorittu (changes requested):
Hi Paula, first of all congratulations for finishing your first API 🎈 I left a comment in the code that point where you missed one of the basic requirement, which was documenting your endpoints for the user to know how to use your api! We recommended you use  Express List Endpoints, it's only few lines of implementation :) the rest looks good and works fine, fix this little thing to pass the assignement, thank you! 🤝

AntonellaMorittu (changes requested):
Hi Paula, I saw you added the endpoint list, great! It's still not showing, have you tested it ? Try triggering a manual deployment :) Thanks

AntonellaMorittu (approved):

PR #520, opened 2024-12-05
Project Express API by Anna

Anna2024WebDev:
## Netlify link
https://project-express-api-cats.onrender.com


README.md
```diff
@@ -1,13 +1,13 @@
...
 
-Start by briefly describing the assignment in a sentence or two. Keep it short and to the point.
+I started out by planning for which endpoints and routes to have and decided to have one filter query parameters and conditionals- if/else statements for all cats and filtered by personality, fur_length, or commonality. And then I wanted separate route parameters for breed and ID and used the find method for that. I also added error handling to all routes.
 
 ## The problem
 
-Describe how you approached to problem, and what tools and techniques you used to solve it. How did you plan? What technologies did you use? If you had more time, what would be next?
+I looked a lot at Matildas live session where she had different examples of query parameters or route parameters and filtering or find methods. I think it was very straight forward but I had one problem with the find method and route paths, there was a conflict between the routes for breed and ID so I needed to add more unique path to the cats breed. "/cats/breed/:breed" instead of "/cats/:breed" which worked.
```
HIPPIEKICK, line 9: The most RESTful thing would've been to put the breed as a query param (`/cats?breed=bengal`) instead of creating its own route. This way, you've still named the endpoint after what it returns: cats (filtered on breed)

server.js
```diff
@@ -1,30 +1,98 @@
...
+  response.json({
+    message: "Miao Miao! Welcome to the Cats API!  Here you can find all types of cats. Below are the available endpoints",
+    endpoints: endpoints // Send the list of endpoints as JSON
+  })
+})
+
+// Combined cats query parameter filtering on all cats and cats with category query filter for personality, fur length and commonality. 
+
+app.get("/cats", (request, response) => {  // http://localhost:9000/cats -route for displaying all cat breeds. 
+  const {personality, fur_length, commonality}= request.query
```
HIPPIEKICK, line 34: Nice! Here you could've added breed as well

server.js
```diff
@@ -1,30 +1,98 @@
...
+  app.get("/cats/:id", (request, response) => {
+    const id = parseInt(request.params.id) // Correctly parse the ID as an integer
+    console.log("Requested ID:", id);  // Log the requested ID
+  
+    const cat = cats.find(cat => cat.id === +id)   //Find specific ID for a cat.
+  if (cat) {
+  response.status(200).json(cat)
+  } else {
+  response.status(404).send("No cat found with that ID")
+  }
```
HIPPIEKICK, line 85: Indentation!

server.js
```diff
@@ -1,30 +1,98 @@
...
+    } else {
+      response.status(404).send("No cats found with the specified criteria") //if no matches when writing a cat personality, fur_length or commonality, this message will show. 
+    }
+  })
+
+ // Route parameters with find method. 
+  app.get ("/cats/breed/:breed", (request, response) => {
+    const breed = request.params.breed.trim().toLowerCase()
+
+    const cat = cats.find (cat => cat.breed.toLowerCase() === breed)  //Find specific cat breed.
```
HIPPIEKICK, line 67: Could you think of a better variable name? 👀

server.js
```diff
@@ -1,30 +1,98 @@
...
+  })
+
+ // Route parameters with find method. 
+  app.get ("/cats/breed/:breed", (request, response) => {
+    const breed = request.params.breed.trim().toLowerCase()
+
+    const cat = cats.find (cat => cat.breed.toLowerCase() === breed)  //Find specific cat breed.  
+    //https://project-express-api-cats.onrender.com/cats/breed/russian%20blue
+  if (cat) {
+    response.status (200).json (cat)
```
HIPPIEKICK, line 70: Nice that you included status codes! But both status and json are JS methods and therefor it should not be a space between method and ()

HIPPIEKICK (approved):
Good job Anna! Just remember to keep it clean and RESTful until next time :)

README.md
Anna2024WebDev: I thought about that first but then it said in the requirements that we should have "A minimum of one endpoint to return a single result (single element)." so then I added it as a route parameter so I had 2 of the single element endpoints.

server.js
Anna2024WebDev: I thought breed was a good name since I only have one key in the json file called breed. But another suggestion could be requestedBreed.

PR #523, opened 2024-12-06
project-express-api - Elina Eriksson Hult

ElinaEH:
https://project-express-api-elina.onrender.com


server.js
```diff
@@ -1,30 +1,80 @@
...
 app.use(express.json());
 
-// Start defining your routes here
+// Endpoint to fetch all books
+app.get("/books", (req, res) => {
+  res.json(booksData);
+});
+
+// Function to filter books by their language
+const filterBooksByLanguage = (req, res) => {
```
HIPPIEKICK, line 22: Nice that you broke this out!

server.js
```diff
@@ -1,30 +1,80 @@
...
+  } else {
+    res.status(404).json({
+      error: `No books found for this language: ${language}`,
+      example_usage: "Try this: /books/language?language=spa"
+    });
+  }
+};
+
+// Endpoint to fetch books filtered by language
+app.get("/books/language", filterBooksByLanguage);
```
HIPPIEKICK, line 52: It would be more RESTful if this was a query param in the /books endpoint instead of its own endpoint, e.g. `/books?language=english`

server.js
```diff
@@ -1,30 +1,80 @@
...
+
+// Root endpoint to list all available API routes dynamically
 app.get("/", (req, res) => {
-  res.send("Hello Technigo!");
+  const endpoints = listEndpoints(app); // Fetch all endpoints
+  res.json({
+    message: "Welcome to the Book API 📚 Here are the available endpoints:",
+    endpoints: endpoints,
+  });
 });
```
HIPPIEKICK, line 75: I usually order my endpoints from the most general to the most specific. Is there a reason why you put this last? 👀

server.js
```diff
@@ -1,30 +1,80 @@
...
+    book => book.language_code.toLowerCase() === language.toLowerCase()
+  );
+
+  // Return the filtered books if matches are found, or show error
+  if (filteredBooks.length > 0) {
+    res.json(filteredBooks);
+  } else {
+    res.status(404).json({
+      error: `No books found for this language: ${language}`,
+      example_usage: "Try this: /books/language?language=spa"
```
HIPPIEKICK, line 46: Very helpful ⭐

HIPPIEKICK (approved):
Good job Elina with your first API! Just remember going forward that all endpoints should be named after what they return and that you can make use of query params

PR #527, opened 2024-12-07
Maryyy_ux_project-express-api

Maryyy-ux:
https://maryyy-ux-project-express-api.onrender.com


server.js
```diff
@@ -1,30 +1,92 @@
...
+  res.json({
+    message: "Welcome to the Books API",
+    routes: [
+      { method: "GET", endpoint: "/books", description: "Get all books" },
+      { method: "GET", endpoint: "/books/:id", description: "Get a single book by ID" },
+      { method: "GET", endpoint: "/books/search", description: "Search books by query parameters" },
+      { method: "GET", endpoint: "/books/page", description: "Get paginated books" },
+    ],
+  });
 });
```
HIPPIEKICK, line 22: Next week you might want to look into Express List Endpoints (see instructions) so that you don't have to update your docs manually when you change something

server.js
```diff
@@ -1,30 +1,92 @@
...
+    return res.status(404).json({ error: "No books found for the given page" });
+  }
+
+  res.json({
+    page: pageNum,
+    limit: limitNum,
+    totalBooks: booksData.length,
+    books: paginatedBooks,
+  });
+});
```
HIPPIEKICK, line 55: This could be a query param under the /books endpoint instead: `/books?page=1`

server.js
```diff
@@ -1,30 +1,92 @@
...
+  const results = booksData.filter((book) =>
+    book.title.toLowerCase().includes(title.toLowerCase()) // Cambié 'books' por 'booksData'
+  );
+
+  if (results.length === 0) {
+    return res.status(404).json({ error: "No books found with the given title" });
+  }
+
+  res.json(results);
+});
```
HIPPIEKICK, line 74: This as well: `/books?search=harry`

HIPPIEKICK (approved):
Good job creating your first API Maria! Just remember going forward to keep it RESTful by using query params instead of creating endpoints for different filters. Endpoints should be named after what they return and you're returning books in all of them

PR #529, opened 2024-12-08
Express API - Dog breeds - Jenny A

jempa182:
## Netlify link
https://project-express-api-y1pr.onrender.com/

### DIFF
--- a/server.js
+++ b/server.js
@@ -1,6 +1,7 @@
+const listEndpoints = require('express-list-endpoints');
 const express = require("express");
 const cors = require("cors");
-const dogs = require("./data/dogs.json");
+const dogs = require("./data/dog-breeds.json");
 
 const port = process.env.PORT || 8080;
 const app = express();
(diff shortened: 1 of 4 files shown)


server.js
```diff
@@ -1,30 +1,76 @@
...
+    );
+  }
+  if (familyFriendly) {
+    filteredDogs = filteredDogs.filter(dog => 
+      dog.familyFriendly.toLowerCase() === familyFriendly.toLowerCase()
+    );
+  }
+  
+  res.json(filteredDogs);
+});
```
HIPPIEKICK, line 62: Excellent work making use of query params and chaining the filters like this, works like a charm ⭐

HIPPIEKICK (approved):
Good job creating a RESTful API Jenny. Looks super! Impressive work with chaining the filters as well as implementing error handling. Just remember to clean up the code when you're done (remove console.logs)

---
This is not part of the review, but what the hell kind of category is this 😂 😭 
Skärmavbild 2024-12-10 kl  14 48 28

server.js
jempa182: Thank you for the feedback @HIPPIEKICK! Hahaha I was also really confused by the "Toy" category so I asked ChatGPT and apparently it's "the official classification for small breeds that were specifically developed to be companion pets rather than working dogs". So small lap dogs! You live and you learn 😂😂

PR #531, opened 2024-12-08
Week 17: API 

jacquelinekellyhunt:
https://golden-globes-api.onrender.com

### DIFF
--- a/routes/awards.js
+++ b/routes/awards.js
@@ -1,0 +1,40 @@
+import express from 'express';
+import awards from '../data/golden-globes.json';
+
+const router = express.Router();
+
+router.get('/', (req, res) => {
+  const { year, category, winner } = req.query;
+  let filteredAwards = awards;
+
+  if (year) {
+    filteredAwards = filteredAwards.filter((award) => award.year_award == year);
+  }
+  if (category) {
+    filteredAwards = filteredAwards.filter((award) =>
+      award.category.toLowerCase().includes(category.toLowerCase())
+    );
+  }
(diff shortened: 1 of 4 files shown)


routes/awards.js
HIPPIEKICK, line 1: Is this file used? 👀

server.js
```diff
@@ -1,30 +1,67 @@
...
+      award.category.toLowerCase().includes(category.toLowerCase())
+    );
+  }
+
+  if (filteredAwards.length === 0) {
+    return res.status(404).json({ error: "No awards found matching the criteria" });
+  }
+
+  res.status(200).json(filteredAwards);
+});
```
HIPPIEKICK, line 52: These endpoints could be one if you make use of another query param, such as `/awards?winner=true`

server.js
```diff
@@ -1,30 +1,67 @@
 import express from "express";
 import cors from "cors";
+import awardsData from "./data/golden-globes.json";
```
HIPPIEKICK, line 3: I don't understand how your API can function when I can't see any data folder in your repo 😅

HIPPIEKICK (approved):
I might be stupid but I'm so confused about the non-existing data file 😅

Remember to make use of query params going forward, instead of creating different endpoints to filter.

server.js
jacquelinekellyhunt: @HIPPIEKICK what do you mean exactly by not seeing a data folder? :'))
Screenshot 2024-12-11 at 11 47 34

PR #532, opened 2024-12-08
Express API - Kpop albums - Cholpon 

cholpoun:
https://project-express-api-fy1c.onrender.com


server-compiled.js
HIPPIEKICK, line 1: You don't need to include the compiled files, you can add it to your gitignore file

server.js
```diff
@@ -1,27 +1,115 @@
...
+
+// Load the dataset
+let data = [];
+try {
+  const rawData = fs.readFileSync(dataPath, "utf-8");
+  data = JSON.parse(rawData);
+  console.log("Dataset loaded successfully:", data.length, "records found.");
+} catch (error) {
+  console.error("Error loading dataset:", error.message);
+}
```
HIPPIEKICK, line 18: Why? 👀

server.js
```diff
@@ -1,27 +1,115 @@
...
+app.get("/api/albums/rating/:rating", (req, res) => {
+  const rating = parseFloat(req.params.rating);
+  const albums = data.filter((album) => album.rating >= rating);
+
+  if (albums.length > 0) {
+    res.json(albums);
+  } else {
+    res.status(404).json({ error: "No albums found with the given rating" });
+  }
 });
```
HIPPIEKICK, line 113: These could all be query params under the /albums endpoint, no need to create different endpoints for each filter. This way you also have the possibility to chain filters, e.g. if you want to filter on year _and_ category

HIPPIEKICK (approved):
Good job creating your first API Cholpon! Just remember going forward to make use of query params instead of creating one endpoint for each filter

PR #535, opened 2024-12-08
Climbing routes api - Zoe

zoe-lindqvist:
## Netlify link
https://project-express-api-mqhs.onrender.com/


index.js
```diff
@@ -0,0 +1,163 @@
...
+      "Welcome to the Climbing Routes API! Here are the available endpoints:",
+    endpoints: listEndpoints(app),
+  });
+});
+
+// Return all climbing routes
+// http://localhost:9000/climbingRoutes
+app.get("/climbingRoutes", (_, res) => {
+  res.json(climbingRoutes);
+});
```
HIPPIEKICK, line 25: Instead of creating different endpoints for each filter/sorting option, you should just include all of the query params under this endpoint. This way you can also chain the filters, if anyone wants to filter on e.g. type _and_ difficulty.

This is good practice to make your API RESTful.

index.js
```diff
@@ -0,0 +1,163 @@
...
+app.get("/climbingRoutes/:id", (req, res) => {
+  const id = req.params.id;
+
+  const climbingRoute = climbingRoutes.find((route) => route.id === +id);
+  if (climbingRoute) {
+    res.json(climbingRoute);
+  } else {
+    res.status(404).send("No climbing route found with that ID");
+  }
+});
```
HIPPIEKICK, line 128: ⭐ Nice with some error handling as well 👍

HIPPIEKICK (approved):
Good job with your first API Zoe! Just remember going forward that endpoints should be named after what they return so a RESTful approach is to include all filter options as query params in your main endpoint, e.g. `/climbing-routes?type=sport&difficulty=low`

PR #536, opened 2024-12-08
Harry Potter express API 

Fannyhenriques:
## Netlify link
https://project-express-api-qyuo.onrender.com/

### DIFF
--- a/server.js
+++ b/server.js
@@ -1,5 +1,6 @@
 import express from "express";
 import cors from "cors";
+import listEndpoints from "express-list-endpoints";
 import harryPotterCharactersData from "./data/harry-potter-characters.json";
 
 // when starting the server. Example command to overwrite PORT env variable value:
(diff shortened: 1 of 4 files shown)


server.js
```diff
@@ -21,10 +14,75 @@ app.use(express.json());
...
+    const charactersIntroducedInYear = harryPotterCharactersData.filter(character =>
+      character.yearIntroduced === +year
+    );
+    res.json(charactersIntroducedInYear);
+  }
+  else {
+    res.json(harryPotterCharactersData);  // No filters = return all characters
+
+  }
+});
```
HIPPIEKICK, line 55: Nice job making use of query params to filter ⭐ Could you think of a way to improve this code so that the user can filter on two things at once? E.g. `/characters?role=student&year=1997`

HIPPIEKICK (approved):
Good job with your first API Fanny! You've met the requirements, but I'd like to challenge you to think of a way to chain your filters 👀

PR #537, opened 2024-12-11
Main

xingyin2024:
## view it live
https://traktamente-express-api.onrender.com


HIPPIEKICK (dismissed):
Have a look at the requirements; you should create at least one endpoint that returns many resources (done) and at least one endpoint that returns a single resource (not done)

xingyin2024:
Hi, i have add new single endpoints of code and country to return one single result which is the first data found in json. Have a look!

JennieDalgren (approved):
⭐

